PR #1, opened 2020-06-05
Add item list CRD Spec

xianlubird:
```
# The list generator specifies a literal list of argument values to the app spec template.
kind: ApplicationSet
metadata:
  name: guestbook
spec:
  generators:
  - list:
      items:
      - cluster: engineering-dev
        url: https://1.2.3.4
      - cluster: engineering-prod
        url: https://2.4.6.8
      - cluster: finance-preprod
        url: https://9.8.7.6
  template:
    metadata:
      name: '{{name}}-guestbook'
    spec:
      source:
        repoURL: https://github.com/infra-team/cluster-deployments.git
        targetRevision: HEAD
        path: guestbook/{{cluster}}
      destination:
        server: '{{url}}'
        namespace: guestbook

```

First crd implement  list item function


Makefile
```diff
@@ -0,0 +1,27 @@
...
+IMAGE_PREFIX?=registry.cn-hangzhou.aliyuncs.com/appcenter
+DOCKER_PUSH?=true
+
+.PHONY: build
+build:
+	CGO_ENABLED=0 go build -ldflags="-w -s" -o ./dist/argocd-aplicationset .
+
+.PHONY: image
+image:
+	docker build -t registry.cn-hangzhou.aliyuncs.com/appcenter/argocd-aplicationset:$(IMAGE_TAG) .
```
dgoodwin, line 12: Could we make the registry configurable as well, $IMAGE perhaps?

jessesuen: I agree. Can make it such that it builds a local image (without a registry), but will prepend `IMAGE_PREFIX` if it is defined?

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              ApplicationSetSpec `json:"spec"`
+}
+
+// ApplicationSetSpec represents a class of application set state.
+type ApplicationSetSpec struct {
+	Generators ApplicationSetGenerators  `json:"generators"`
+	Template   ApplicationSetTemplate    `json:"template"`
+	Operation  *v1alpha1.SyncOperation   `json:"operation,omitempty"`
```
jessesuen, line 23: Regarding the `operation` field:

1. Reusing the SyncOperation from the Applications APIs probably doesn't make sense here, since that object deals with a single git revision. But ApplicationSets may deal with many revisions, and even many repositories. If we do decide to have operation, we will need a new data structure.

2. In fact, we're not sure what a sync operation even means for an AppSet. Currently, I think it would only be used as a convenience field which carries forward setting the operation field for all the Applications in the ApplicationSet. But this could be done as a convenience in the API server instead.

3. We've had some regrets about introducing this operation field in Applications, since a better alternative might have been to introduce an `SyncOperation` CRD and controller to perform the logic. This CRD would act like a Job object and finish Successfully or Failed. 

So with the above, I think we should leave out operation, until we think about how we want to do operations (if at all).

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+type ApplicationSetSyncPolicy struct {
+	// Automated will keep an application synced to the target revision
+	Automated *SyncPolicyAutomated `json:"automated,omitempty"`
+}
+
+// SyncPolicyAutomated
+type SyncPolicyAutomated struct {
+	// Prune will prune resources automatically as part of automated sync (default: false)
+	Prune       bool `json:"prune,omitempty"`
+	InitialSync bool `json:"initialSync,omitempty"`
```
jessesuen, line 37: I'm thinking the word `sync` is redundant in the `initialSync`, since it's already part of a `syncPolicy`. How about:

```yaml
spec:
  syncPolicy:
    automated:
      prune: true
      initial: true
```

Let's also comment this to explain what this option is for:
```
// Initial will perform an initial sync for any newly created Applications which do not have automated sync turned on.
```

api/v1alpha1/applicationset_types.go
xianlubird: agree

api/v1alpha1/applicationset_types.go
xianlubird: Done

Makefile
xianlubird: Done

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+	TemplateSpec      v1alpha1.ApplicationSpec `json:"spec"`
+}
+
+// ApplicationSetGenerators include list item info
+type ApplicationSetGenerators struct {
+	List GeneratorsList `json:"list, omitempty"`
+}
+
+// GeneratorsList include items info
+type GeneratorsList struct {
```
dgoodwin, line 52: GeneratorsList implies this is a list of all generators, what do you think about ListGenerator to more clearly indicate what this is?

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+	Items []GeneratorsItems `json:"items"`
+}
+
+// GeneratorsItems include cluster and url info
+type GeneratorsItems struct {
```
dgoodwin, line 57: Suggest ListGeneratorItem

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+
+// ApplicationSetTemplate represents argocd ApplicationSpec
+type ApplicationSetTemplate struct {
+	metav1.ObjectMeta `json:"metadata"`
+	TemplateSpec      v1alpha1.ApplicationSpec `json:"spec"`
+}
+
+// ApplicationSetGenerators include list item info
+type ApplicationSetGenerators struct {
+	List GeneratorsList `json:"list, omitempty"`
```
dgoodwin, line 48: Feels like this should be *GeneratorsList as it is optional.

main.go
```diff
@@ -0,0 +1,72 @@
...
+		LeaderElectionID:   "58ac56fa.",
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err = (&controllers.ApplicationSetReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("ApplicationSet"),
```
dgoodwin, line 58: Minor compile error here as the reconciler has no Log field.

api/v1alpha1/applicationset_types.go
xianlubird: Agree

api/v1alpha1/applicationset_types.go
xianlubird: Yes

main.go
xianlubird: fix

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type ApplicationSet struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              ApplicationSetSpec `json:"spec"`
+}
+
+// ApplicationSetSpec represents a class of application set state.
+type ApplicationSetSpec struct {
+	Generators ApplicationSetGenerators  `json:"generators"`
```
jessesuen, line 21: Per spec, it should be an array of ApplicationSetGenerator:
```yaml
spec:
  generators:
  - clusters: {}
```

```golang
    Generators []ApplicationSetGenerator  `json:"generators"`
```

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// ApplicationSetTemplate represents argocd ApplicationSpec
+type ApplicationSetTemplate struct {
+	metav1.ObjectMeta `json:"metadata"`
+	TemplateSpec      v1alpha1.ApplicationSpec `json:"spec"`
+}
+
+// ApplicationSetGenerators include list item info
+type ApplicationSetGenerators struct {
```
jessesuen, line 47: We should remove `s` from ApplicationSetGenerators since this should be a single generator

```golang
type ApplicationSetGenerator struct {
	List *ListGenerator `json:"list, omitempty"`
}
```

Also, we should remove s from `ListGenerators` as well.

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+	List *ListGenerators `json:"list, omitempty"`
+}
+
+// ListGenerators include items info
+type ListGenerators struct {
+	Items []ItemGenerators `json:"items"`
+}
+
+// ItemGenerators include cluster and url info
+type ItemGenerators struct {
```
jessesuen, line 57:
```golang
type ListGeneratorItem struct
```

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+	TemplateSpec      v1alpha1.ApplicationSpec `json:"spec"`
+}
+
+// ApplicationSetGenerators include list item info
+type ApplicationSetGenerators struct {
+	List *ListGenerators `json:"list, omitempty"`
+}
+
+// ListGenerators include items info
+type ListGenerators struct {
```
jessesuen, line 52: Should be:
```golang
type ListGenerator struct {
```

(no s)

api/v1alpha1/applicationset_types.go
```diff
@@ -0,0 +1,73 @@
...
+	// Prune will prune resources automatically as part of automated sync (default: false)
+	Prune bool `json:"prune,omitempty"`
+	// Initial will perform an initial sync for any newly created Applications which do not have automated sync turned on.
+	Initial bool `json:"initial,omitempty"`
+}
+
+// ApplicationSetTemplate represents argocd ApplicationSpec
+type ApplicationSetTemplate struct {
+	metav1.ObjectMeta `json:"metadata"`
+	TemplateSpec      v1alpha1.ApplicationSpec `json:"spec"`
```
jessesuen, line 43: I think the field name needs to be named `Spec`, otherwise this may be an K8s API violation.

PR #2, opened 2020-06-11
Sketch out cluster generator in the CRD, and generator interface

dgoodwin:
Builds on #1 to add a field for cluster generator with label selector.

Fixes the main.go compile problem, and adds dist/ to .gitignore.

Adds a simple interface for generators and two empty implementations for list and cluster.


pkg/generators/interface.go
```diff
@@ -0,0 +1,14 @@
...
+	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
+)
+
+// Generator defines the interface implemented by all ApplicationSet generators.
+type Generator interface {
+	// GenerateApplications interprets the ApplicationSet and generates all relevant Applications.
+	// The expected / desired list of Applications is returned, it then needs to be reconciled
+	// against the current state of the Applications in the cluster.
+	GenerateApplications(appSet *argoprojiov1alpha1.ApplicationSet) ([]argov1alpha1.Application, error)
+}
```
dgoodwin, line 14: I thought it might be helpful to break this into an interface and implement for each type of generator, to help us work without causing git conflicts as much as possible.

Returns the Applications (desired state) to help with unit testing, something else will need to reconcile these with the actual state of the Applications currently.

Common code can be shared amongst generators using functions in this package, but my suspicion is the implementations of each generator will be unique enough to isolate them as separate implementations.

main.go
```diff
@@ -0,0 +1,72 @@
...
+		LeaderElectionID:       "58ac56fa.",
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err = (&controllers.ApplicationSetReconciler{
+		Client: mgr.GetClient(),
+		//Log:    ctrl.Log.WithName("controllers").WithName("ApplicationSet"),
```
dgoodwin, line 58: Temporarily commented to avoid the compile error.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN go mod download
+
+# Copy the go source
+COPY . .
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o applicationset-controller main.go
+
+# Use distroless as minimal base image to package the manager binary
+FROM debian:10-slim
```
OmerKahani, line 19: can we use alpine?

xianlubird: of course.

I copy this from argo-cd Dockerfile.

I think it's ok to use alpine.

main.go
xianlubird: I have fix this

Dockerfile
jessesuen: Better yet would be `scratch`.

Also, we should run as a non-root, numeric user (i.e. 999).
```Dockerfile
USER 999
```

See https://github.com/argoproj/argo-cd/pull/3108 for rationale.

dgoodwin:
Rebased onto latest changed and ready for review.

pkg/generators/list.go
```diff
@@ -0,0 +1,29 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ Generator = (*ListGenerator)(nil)
+
+type ListGenerator struct {
+	commonClient client.Client
+}
+
+func NewListGenerator(client client.Client) Generator {
```
jessesuen, line 16: I don't think the list generator should need a kubernetes client, given the items are a hard-wired list.

pkg/generators/cluster.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var _ Generator = (*ClusterGenerator)(nil)
+
+// ClusterGenerator generates Applications for some or all clusters registered with ArgoCD.
+type ClusterGenerator struct {
+	commonClient client.Client
+}
+
+func NewClusterGenerator(client client.Client) Generator {
```
jessesuen, line 17: Instead of a client, can we accept an informer? This way we can avoid API calls and get the values from the informer cache.

dgoodwin: I believe the controller-runtime client is a delegating client and an abstraction over the informers, so reads via the client are not hitting the API server. (ref https://github.com/kubernetes-sigs/controller-runtime/issues/498) Does that seem ok? If not we can go informer. I think I'll actually just remove it for now as I'm not totally sure how to get access to the informers underneath but will handle that once I get going on impl if that is the preferred direction.

pkg/generators/list.go
dgoodwin: Removed.

pkg/generators/list.go
```diff
@@ -0,0 +1,29 @@
...
+	g := &ListGenerator{
+		commonClient: client,
+	}
+	return g
+}
+
+func (g *ListGenerator) GenerateApplications(appSet *argoprojiov1alpha1.ApplicationSet) ([]argov1alpha1.Application, error) {
+	if appSet == nil {
+		return nil, fmt.Errorf("ApplicationSet is empty ")
```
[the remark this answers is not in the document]
dgoodwin: Fixed in both implementations.

pkg/generators/cluster.go
xianlubird: I think the  controller-runtime client is fine

xianlubird:
What's your opinion about this pr @jessesuen

pkg/generators/cluster.go
jessesuen: Thanks for clarifying. I think you're right. Haven't used kubebuilder before so TIL.

PR #7, opened 2020-07-03
Add git directory generator

OmerKahani:
closes #12

PR #8, opened 2020-07-03
Cluster Generator reconciling on cluster secret events

dgoodwin:
This PR enables the cluster generator to run and locate all cluster secrets with matching labels. The controller now watches for changes to cluster Secrets and reconciles all appsets with a Cluster generator.

### DIFF
--- a/examples/clusters.yaml
+++ b/examples/clusters.yaml
@@ -10,13 +10,17 @@ metadata:
   name: guestbook
 spec:
   generators:
-  - clusters: {}
+  - clusters:
+      selector:
+        matchLabels:
+          argocd.argoproj.io/secret-type: cluster
   template:
     metadata:
       name: '{{name}}-guestbook'
       labels:
         environment: '{{metadata.labels.environment}}'
     spec:
+      project: ""
       source:
         repoURL: https://github.com/infra-team/cluster-deployments.git
         targetRevision: HEAD
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	argoprojiov1alpha1 "github.com/argoproj-labs/applicationset/api/v1alpha1"
+	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
 )
 
 // ApplicationSetReconciler reconciles a ApplicationSet object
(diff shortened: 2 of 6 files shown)


examples/clusters.yaml
```diff
@@ -10,13 +10,17 @@ metadata:
   name: guestbook
 spec:
   generators:
-  - clusters: {}
+  - clusters:
+      selector:
+        matchLabels:
+          argocd.argoproj.io/secret-type: cluster
```
dgoodwin, line 16: This example just matches *every* cluster argo knows about as we're re-using the argo label.

pkg/controllers/applicationset_controller.go
```diff
@@ -49,14 +54,24 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	var generator generators.Generator
-	generator = generators.NewListGenerator()
+	listGenerator := generators.NewListGenerator()
+	clusterGenerator := generators.NewClusterGenerator(r.Client)
```
dgoodwin, line 58: This portion will conflict with #7 but I can fix up once it merges.

pkg/controllers/applicationset_controller.go
```diff
@@ -49,14 +54,24 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
+
+	// desiredApplications is the main list of all expected Applications from all generators in this appset.
+	var desiredApplications  []argov1alpha1.Application
 	for _, tmpGenerator := range applicationSetInfo.Spec.Generators {
-		desiredApplications, err := generator.GenerateApplications(&tmpGenerator, &applicationSetInfo)
-		log.Infof("desiredApplications %+v", desiredApplications)
+		var apps []argov1alpha1.Application
```
xianlubird, line 63: If we have many kinds of generator,  will this increase many `if else` struct .

should we provide a abstract function to hide these like factory.

dgoodwin: Per slack it sounds like perhaps we should instantiate all generators we know of, then call each with the AppSet, and let it return us the list of Applications for every item in the list that matches.

Although we should probably discuss if it makes sense to have multiple generators in one AppSet at some point, I'm a little worried that'll never be used due to the nature of the templating and the available params being different.

Should we defer this and followup PR once we know what to do?

xianlubird: yes.  fine

PR #11, opened 2020-07-11
Update

OmerKahani:
Refactor the create application method so it will also do update and delete.

Not sure about the name of the function, does doing delete in an apply function is ok?

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -125,6 +125,7 @@ func (r *ApplicationSetReconciler) createOrUpdateInCluster(ctx context.Context, 
 
 	//create or updates the application in appList
 	for _, app := range desiredApplications {
+		appLog := log.WithFields(log.Fields{"app": app.Name, "appSet": applicationSet.Name})
 		app.Namespace = applicationSet.Namespace
 
 		found := app
@@ -134,12 +135,12 @@ func (r *ApplicationSetReconciler) createOrUpdateInCluster(ctx context.Context, 
 		})
 
 		if err != nil {
-			log.Error(err, fmt.Sprintf("failed to CreateOrUpdate Application %s resource for applicationSet %s", app.Name, applicationSet.Name))
+			appLog.WithError(err).Errorf("failed to %s Application", action)
 			continue
 		}
 
 		r.Recorder.Eventf(&applicationSet, core.EventTypeNormal, fmt.Sprint(action), "%s Application %q", action, app.Name)
-		log.Infof("%s Application %s resource for applicationSet %s", action, app.Name, applicationSet.Name)
+		appLog.Logf(log.InfoLevel, "%s Application", action)
 	}
 }
 
@@ -159,16 +160,17 @@ func (r *ApplicationSetReconciler) deleteInCluster(ctx context.Context, applicat
 
 	// Delete apps that are not in m[string]bool
 	for _, app := range current.Items {
+		appLog := log.WithFields(log.Fields{"app": app.Name, "appSet": applicationSet.Name})
 		_, exists := m[app.Name]
 
 		if exists == false {
 			err := r.Client.Delete(ctx, &app)
 			if err != nil {
-				log.Error(err, fmt.Sprintf("failed to delete Application %s resource for applicationSet %s", app.Name, applicationSet.Name))
+				appLog.WithError(err).Error("failed to delete Application")
 				continue
 			}
 			r.Recorder.Eventf(&applicationSet, core.EventTypeNormal, "Deleted", "Deleted Application %q", app.Name)
-			log.Infof("Deleted Application %s resource for applicationSet %s", app.Name, applicationSet.Name)
+			appLog.Log(log.InfoLevel, "Deleted application")
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/controllers/applicationset_controller.go
```diff
@@ -111,21 +145,30 @@ func (r *ApplicationSetReconciler) createApplications(ctx context.Context, appli
 		})
 
 		if err != nil {
-			log.Error(err, fmt.Sprintf("failed to get Application %s resource for applicationSet %s", app.Name, applicationSetInfo.Name))
+			log.Error(err, fmt.Sprintf("failed to CreateOrUpdate Application %s resource for applicationSet %s", app.Name, applicationSetInfo.Name))
```
xianlubird, line 148: We should add `action` in the log

pkg/controllers/applicationset_controller.go
```diff
@@ -111,21 +145,30 @@ func (r *ApplicationSetReconciler) createApplications(ctx context.Context, appli
...
+		r.Recorder.Eventf(&applicationSetInfo, core.EventTypeNormal, fmt.Sprint(action), "%s Application %q", action, app.Name)
 		log.Infof("%s Application %s resource for applicationSet %s", action, app.Name, applicationSetInfo.Name)
 	}
 
-	return nil
-
-}
-
-func (r *ApplicationSetReconciler) getApplications(ctx context.Context, applicationSetInfo argoprojiov1alpha1.ApplicationSet) ([]argov1alpha1.Application, error) {
+	// Delete apps that are not in m[string]bool
```
xianlubird, line 156: I think we should separate out this function ,   delete application shouldn't exists in apply function

OmerKahani: done

pkg/controllers/applicationset_controller.go
```diff
@@ -78,17 +79,34 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 
 	}
 
-	if err := r.createApplications(ctx, applicationSetInfo, desiredApplications); err != nil {
-		log.Infof("Unable to create applications applicationSetInfo %e", err)
-		return ctrl.Result{}, err
-	}
+	r.createOrUpdateInCluster(ctx, applicationSetInfo, desiredApplications)
+	r.deleteInCluster(ctx, applicationSetInfo, desiredApplications)
```
xianlubird, line 83: Should we have some `if conditions`, like if `finalizers` is set , we should delete this cluster?

dgoodwin: Not sure if this is what you meant, we definitely will want to add a finalizer to ApplicationSets, and if deleting the appset we must cleanup all Applications before removing that finalizer. Deletion cleanup should probably be a separate PR. I can try to find some time for this one this week.

pkg/controllers/applicationset_controller.go
OmerKahani: done

pkg/controllers/applicationset_controller.go
OmerKahani: this function deletes old applications from the cluster (deleted record from a list)

Deletion of ApplicationSets is implemented with owner reference so the Kubernetes GC will catch it and delete the applications

pkg/controllers/applicationset_controller.go
```diff
@@ -99,34 +117,60 @@ func (r *ApplicationSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		found := app
 		action, err := ctrl.CreateOrUpdate(ctx, r.Client, &found, func() error {
 			found.Spec = app.Spec
-			return controllerutil.SetControllerReference(&applicationSetInfo, &found, r.Scheme)
+			return controllerutil.SetControllerReference(&applicationSet, &found, r.Scheme)
 		})
 
 		if err != nil {
-			log.Error(err, fmt.Sprintf("failed to get Application %s resource for applicationSet %s", app.Name, applicationSetInfo.Name))
+			log.Error(err, fmt.Sprintf("failed to CreateOrUpdate Application %s resource for applicationSet %s", app.Name, applicationSet.Name))
```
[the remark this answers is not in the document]
OmerKahani: added thanks!

pkg/controllers/applicationset_controller.go
OmerKahani: going over the proposal, every application should have `resources-finalizer.argocd.argoproj.io` entry in `metadata.finalizers` 

Do we also want to support the option to delete a record from a list / directory from git, and not run the finalizers in the application (something like `kubectl delete app NAME --cascade=false`) ?

xianlubird: > Do we also want to support the option to delete a record from a list / directory from git, and not run the finalizers in the application (something like kubectl delete app NAME --cascade=false) ?

I think we don't need to implement this for now

dgoodwin: > this function deletes old applications from the cluster (deleted record from a list)
> 
> Deletion of ApplicationSets is implemented with owner reference so the Kubernetes GC will catch it and delete the applications

My mistake, forgot about that!

dgoodwin:
LGTM. I'll merge it in.

PR #18, opened 2020-08-19
Fix ClusterGenerator

rtnpro:
This implements generating Argo `Application`s from `ClusterGenerator`.

**TODO**
- [x] unit tests

### DIFF
--- a/pkg/generators/cluster.go
+++ b/pkg/generators/cluster.go
@@ -9,8 +9,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
 	argoprojiov1alpha1 "github.com/argoproj-labs/applicationset/api/v1alpha1"
+	"github.com/argoproj-labs/applicationset/pkg/utils"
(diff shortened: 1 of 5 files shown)


xianlubird:
@OmerKahani Pls take a look

pkg/generators/cluster.go
```diff
@@ -56,9 +57,27 @@ func (g *ClusterGenerator) GenerateApplications(
...
+		params["name"] = cluster.Name
+		params["server"] = string(cluster.Data["server"])
+		for key, value := range cluster.ObjectMeta.Labels {
+			params[fmt.Sprintf("metadata.labels.%s", key)] = value
+		}
 		log.WithField("cluster", cluster.Name).Info("matched cluster secret")
+		tmpApplication, err := utils.RenderTemplateParams(&tmplApplication, params)
+		if err != nil {
+			return nil, err
+		}
```
OmerKahani, line 78: In the list generator, we only log this kind of error
@xianlubird - do we want to stop if there is a problem with ane application or just log it and continue? (I prefer log and continue)

OmerKahani:
Tested with the example, everything is working. @rtnpro Thanks for contributing!

I would love to see unit tests (we are missing them all over the project)

rtnpro:
Thanks for the review @OmerKahani . I will be adding the unitttests for the `ClusterGenerator` tomorrow.

PR #16, opened 2020-08-14
Fix ListGenerator

martin-ducar-gd:
After some testing the ListGenerator (implemented nearly completely). Fixing it to the point of actually generating apps. Also fixing some manifest bugs found in the process

PR #21, opened 2020-08-31
fix(list-generator): return generated applications

mgoodness:
This is the same fix submitted in #16 since @ martin-ducar-gd is blocked by CLA.


dgoodwin (approved):
LGTM

PR #26, opened 2020-09-01
Refactor - Making testing & writing generators more easy

OmerKahani:
1. Extract common parts from the generators. Now a generator only creates the params for the application, and the rendering is happing outside.
Generator now has only one unique responsibility.

2. Refactor the if-else generator chooser to a simpler loop, and move the generators list to outside the function.
Making the code more clean, easy to test, and easy to add a new generator

This PR is going to brake every PR, so it will be better to merge: #25 , #24 , #21 before this one

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -44,6 +44,7 @@ type ApplicationSetReconciler struct {
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
 	Generators []generators.Generator
+	utils.Renderer
 }
 
 // +kubebuilder:rbac:groups=argoproj.io,resources=applicationsets,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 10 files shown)


pkg/generators/interface.go
```diff
@@ -2,14 +2,21 @@ package generators
...
+	GenerateParams(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error)
 }
+
+type EmptyAppSetGeneratorError struct {
+
+}
+
+func (e EmptyAppSetGeneratorError) Error() string{
+	return "ApplicationSet is empty"
+}
```
dgoodwin, line 21: A little more succinct to just do this as:

```
var EmptyAppSetGeneratorError = errors.New("ApplicationSet is empty")
```

pkg/generators/interface.go
```diff
@@ -2,14 +2,21 @@ package generators
...
 )
 
 // Generator defines the interface implemented by all ApplicationSet generators.
 type Generator interface {
 	// GenerateApplications interprets the ApplicationSet and generates all relevant Applications.
 	// The expected / desired list of Applications is returned, it then needs to be reconciled
 	// against the current state of the Applications in the cluster.
-	GenerateApplications(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator,
-		appSet *argoprojiov1alpha1.ApplicationSet) ([]argov1alpha1.Application, error)
+	GenerateParams(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error)
```
dgoodwin, line 12: Godoc needs updating here.

pkg/controllers/applicationset_controller.go
```diff
@@ -61,39 +61,55 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
+func getTempApplication(applicationSetTemplate argoprojiov1alpha1.ApplicationSetTemplate) *argov1alpha1.Application{
+	var tmplApplication argov1alpha1.Application
+	tmplApplication.Namespace = applicationSetTemplate.Namespace
+	tmplApplication.Name = applicationSetTemplate.Name
+	tmplApplication.Spec = applicationSetTemplate.Spec
+
+	return &tmplApplication
+}
+
+func (r *ApplicationSetReconciler) extractApplications(applicationSetInfo argoprojiov1alpha1.ApplicationSet) []argov1alpha1.Application {
```
dgoodwin, line 82: generateApplications seems like a more clear name for this function to me.

pkg/controllers/applicationset_controller_test.go
```diff
@@ -16,7 +19,149 @@ import (
...
+
+			generatorMock := generatorMock{}
+			generator := argoprojiov1alpha1.ApplicationSetGenerator{
+				List: &argoprojiov1alpha1.ListGenerator{},
+			}
+
+			generatorMock.On("GenerateParams", &generator).
+				Return(cc.params, cc.generateParamsError)
+
+			rendererMock := rendererMock{}
```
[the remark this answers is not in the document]
OmerKahani: I agree we need to test the RenderTemplateParams, with unit tests or E2E; but I don't think that the applicationset_controller should test it, it's not part of the application_controller responsbilty

pkg/controllers/applicationset_controller_test.go
```diff
@@ -16,7 +19,149 @@ import (
...
+					}
+				}
+			}
+
+			r := ApplicationSetReconciler{
+				Client:   client,
+				Scheme:   scheme,
+				Recorder: record.NewFakeRecorder(1),
+				Generators: []generators.Generator{
+					&generatorMock,
```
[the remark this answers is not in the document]
OmerKahani: I think this is more E2E test, I will think about how to do it and open an issue

pkg/generators/interface.go
OmerKahani: fixed the Godoc

xianlubird (approved):

PR #37, opened 2020-09-04
Improve and document the developer deployment process.

dgoodwin:
Rather than require developers to change on disk files in the git index,
or create custom patching solutions, introduce a make target that allows
us to build, push, and deploy manifests in a single command.

Adds namespace to the ServiceAccount subject in the ClusterRoleBinding,
as far as I can tell this was invalid yaml.


mgoodness (changes requested):

dgoodwin:
PR updated, thanks!

mgoodness (approved):
LGTM

xianlubird (approved):

PR #38, opened 2020-09-04
Fix rbac error listing Secrets at cluster scope.

dgoodwin:
The Secret Watch established in the ApplicationSet controller is global,
and rbac for global secrets was recently removed resulting in a broken
controller.

To work past this and avoid re-introducing global Secret watch RBAC, the
Manager's cache is now limited to only the namespace in which it is
running. This implies that any attempts to use the client outside that
namespace will not work. However this should be safe as we work with
Applications and Secrets, both of which should not exist beyond the
ArgoCD namespace where we're running.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"flag"
 	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+
 	"os"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -19,6 +21,10 @@ import (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
+)
+
+const (
+	defaultNamespace = "argocd"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


mgoodness:
Cluster-wide Secret access was restored in #21, but I agree that this is a far better approach.

dgoodwin:
My mistake I forgot to update branch today I think. I will push an addition here to drop global Secret rbac, sound ok?

dgoodwin:
Actually do we have any need for a cluster-install at all? I believe Argo does that so it can sync resource to the local cluster, however we don't need to do that. I think a namespace install may be all we need. Does that seem right?

mgoodness:
> Actually do we have any need for a cluster-install at all? I believe Argo does that so it can sync resource to the local cluster, however we don't need to do that. I think a namespace install may be all we need. Does that seem right?

You may very well be correct. I'll look into this today.

main.go
```diff
@@ -21,6 +23,10 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+const (
+	defaultNamespace = "argocd"
```
xianlubird, line 27: another namespace is better ?

main.go
```diff
@@ -43,9 +49,23 @@ func main() {
...
 
+	// Determine the namespace we're running in. Normally injected into the pod as an env
+	// var via the Kube downward API configured in the Deployment. We also default to "argocd"
+	// for developers running the binary locally who may not remember to set a NAMESPACE environment
+	// variable.
+	ns := os.Getenv("NAMESPACE")
+	if len(ns) == 0 {
+		ns = defaultNamespace
+	}
+	setupLog.Info("using argocd namespace", "namespace", ns)
```
xianlubird, line 60: Maybe application set is better

dgoodwin: If we're destined to be incorporated in argocd it might be best to use the namespace where we expect to land. Although I'm not sure if argocd assumes anything about the namespace it's running in...

dgoodwin: Because we're unsure, why don't we remove the concept of a default namespace. If there is no NAMESPACE env var, we error out and ask you to set it, which would only affect developers running the binary locally on their systems. The pod definition will always ensure it's set to whatever the current namespace is.

dgoodwin: Pushed as a separate commit for your consideration.

xianlubird (approved):

PR #39, opened 2020-09-04
Fix missing Application RBAC.

dgoodwin:
We were missing several permissions for Applications which are one of
the primary resources we need to work with.


mgoodness:
Think your branch needs a rebase; the ClusterRole in `master` already grants the necessary permissions for `applications`.

dgoodwin:
My mistake!

PR #42, opened 2020-09-04
Improve controller error handling and logging.

dgoodwin:
Controller was ignoring errors creating/updating/deleting Applications
when it should be returning that error so the object is requeued with
backoff.

Fixed some inaccurate or messy logging.

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -57,15 +57,26 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 
 	var applicationSetInfo argoprojiov1alpha1.ApplicationSet
 	if err := r.Get(ctx, req.NamespacedName, &applicationSetInfo); err != nil {
-		log.Infof("Unable to fetch applicationSetInfo %e", err)
+		if client.IgnoreNotFound(err) != nil {
+			log.WithField("request", req).WithError(err).Infof("unable to get ApplicationSet")
+		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	// desiredApplications is the main list of all expected Applications from all generators in this appset.
-	desiredApplications := r.generateApplications(applicationSetInfo)
+	desiredApplications, err := r.generateApplications(applicationSetInfo)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 
-	r.createOrUpdateInCluster(ctx, applicationSetInfo, desiredApplications)
-	r.deleteInCluster(ctx, applicationSetInfo, desiredApplications)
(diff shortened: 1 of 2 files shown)


pkg/controllers/applicationset_controller.go
```diff
@@ -135,18 +144,19 @@ func (r *ApplicationSetReconciler) createOrUpdateInCluster(ctx context.Context,
 		})
 
 		if err != nil {
-			appLog.WithError(err).Errorf("failed to %s Application", action)
+			appLog.WithError(err).WithField("action", action).Error("failed to create/update Application")
 			continue
```
OmerKahani, line 148: Maybe add a flag to indicate if there were errors, and return the flag at the end of the function

pkg/controllers/applicationset_controller.go
```diff
@@ -167,12 +177,13 @@ func (r *ApplicationSetReconciler) deleteInCluster(ctx context.Context, applicat
 			err := r.Client.Delete(ctx, &app)
 			if err != nil {
 				appLog.WithError(err).Error("failed to delete Application")
-				continue
+				return err
```
OmerKahani, line 180: Instead of stoping the loop, we can try to do the best effort, and continue to delete other applications. We can use a boolean flag to indicate if there was an error

OmerKahani (changes requested):

pkg/controllers/applicationset_controller.go
xianlubird: agree

xianlubird:
Pls fix the conflict

pkg/controllers/applicationset_controller.go
dgoodwin: Done.

pkg/controllers/applicationset_controller.go
dgoodwin: Done. 

It seems to make sense here. We should be careful with the concept of continuing after an error though as there could be some implications, if we applied the same concept up a level and decided to proceed with cleanup of obsolete apps after a failure creating a new one, something unintended could happen.

pkg/controllers/applicationset_controller_test.go
```diff
@@ -57,6 +57,7 @@ func TestExtractApplications(t *testing.T) {
 		template			argoprojiov1alpha1.ApplicationSetTemplate
 		generateParamsError	error
 		rendererError		error
+		expectErr bool
```
dgoodwin, line 60: Per these changes we now expect these tests to do something differently, as we don't want to swallow an error, we want to return it and let the queue retry with backoff.

dgoodwin:
PR updated, conflict resolved, tests passing.

xianlubird (approved):

PR #43, opened 2020-09-04
chore(kustomize): set default namespace to argocd

mgoodness:
Sets the default kustomize installation namespace to `argocd`. This can be overridden by end-users in their own overlays:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: foo

resources:
  - github.com/argoproj-labs/applicationset/manifests/cluster-install
```


dgoodwin (approved):
LGTM

PR #45, opened 2020-09-09
Implement Git files discovery

### LINKED ISSUE #29: Implement Git Files Discovery
Implement this example https://github.com/argoproj-labs/applicationset/blob/master/examples/git-files-discovery.yaml

Relate to #28 

Might requires changes to the repo service

rtnpro:
Fixes #29

### DIFF
--- a/pkg/generators/git.go
+++ b/pkg/generators/git.go
@@ -76,18 +76,20 @@ func (g *GitGenerator) generateParamsForGitDirectories(appSetGenerator *argoproj
 }
 
 func (g *GitGenerator) generateParamsForGitFiles(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error) {
-	allPaths := []string{}
+	allPaths := make(map[string]bool)
 	for _, requestedPath := range appSetGenerator.Git.Files {
 		paths, err := g.repos.GetPaths(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, requestedPath.Path)
 		if err != nil {
 			return nil, err
 		}
-		allPaths = append(allPaths, paths...)
+		for _, path := range paths {
+			allPaths[path] = true
+		}
 	}
 
 	res := []map[string]string{}
 
-	for _, path := range allPaths {
+	for path, _ := range allPaths {
 		params, err := g.generateParamsFromGitFile(appSetGenerator, path)
 		if err != nil {
 			return nil, err
--- a/pkg/generators/git.go
+++ b/pkg/generators/git.go
@@ -2,11 +2,13 @@ package generators
 
 import (
 	"context"
+	"encoding/json"
 	"path"
 	"time"
 
 	argoprojiov1alpha1 "github.com/argoproj-labs/applicationset/api/v1alpha1"
 	"github.com/argoproj-labs/applicationset/pkg/services"
+	"github.com/jeremywohl/flatten"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -37,6 +39,23 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
 		return nil, EmptyAppSetGeneratorError
 	}
 
+	var err error
+	var res []map[string]string
+	if appSetGenerator.Git.Directories != nil {
+		res, err = g.generateParamsForGitDirectories(appSetGenerator)
+	} else if appSetGenerator.Git.Files != nil {
+		res, err = g.generateParamsForGitFiles(appSetGenerator)
+	} else {
+		return nil, EmptyAppSetGeneratorError
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func (g *GitGenerator) generateParamsForGitDirectories(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error) {
 	allApps, err := g.repos.GetApps(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision)
 	if err != nil {
 		return nil, err
@@ -49,14 +68,52 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
 		"revision": appSetGenerator.Git.Revision,
 	}).Info("applications result from the repo service")
 
-	requestedApps := g.filter(appSetGenerator.Git.Directories, allApps)
+	requestedApps := g.filterApps(appSetGenerator.Git.Directories, allApps)
 
-	res := g.generateParams(requestedApps, appSetGenerator)
+	res := g.generateParamsFromApps(requestedApps, appSetGenerator)
 
 	return res, nil
 }
 
-func (g *GitGenerator) filter(Directories []argoprojiov1alpha1.GitDirectoryGeneratorItem, allApps []string) []string {
+func (g *GitGenerator) generateParamsForGitFiles(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error) {
+	allPaths := []string{}
+	for _, requestedPath := range appSetGenerator.Git.Files {
+		paths, err := g.repos.GetPaths(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, requestedPath.Path)
+		if err != nil {
+			return nil, err
+		}
+		allPaths = append(allPaths, paths...)
+	}
+
+	res := []map[string]string{}
+
+	for _, path := range allPaths {
+		content, err := g.repos.GetFileContent(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, path)
+		if err != nil {
+			return nil, err
+		}
+
+		config := make(map[string]interface{})
+		err = json.Unmarshal(content, &config)
+		if err != nil {
+			return nil, err
+		}
+
+		flat, err := flatten.Flatten(config, "", flatten.DotStyle)
+		if err != nil {
+			return nil, err
+		}
+		params := make(map[string]string)
+		for k, v := range flat {
+			params[k] = v.(string)
+		}
+
+		res = append(res, params)
+	}
+	return res, nil
+}
+
+func (g *GitGenerator) filterApps(Directories []argoprojiov1alpha1.GitDirectoryGeneratorItem, allApps []string) []string {
 	res := []string{}
 	for _, requestedPath := range Directories {
 		for _, appPath := range allApps {
@@ -74,7 +131,7 @@ func (g *GitGenerator) filter(Directories []argoprojiov1alpha1.GitDirectoryGener
 	return res
 }
 
-func (g *GitGenerator) generateParams(requestedApps []string, _ *argoprojiov1alpha1.ApplicationSetGenerator) []map[string]string {
+func (g *GitGenerator) generateParamsFromApps(requestedApps []string, _ *argoprojiov1alpha1.ApplicationSetGenerator) []map[string]string {
 	// TODO: At some point, the appicationSetGenerator param should be used
 
 	res := make([]map[string]string, len(requestedApps))
--- a/pkg/generators/git.go
+++ b/pkg/generators/git.go
@@ -15,10 +15,10 @@ import (
 var _ Generator = (*GitGenerator)(nil)
 
 type GitGenerator struct {
-	repos services.Apps
+	repos services.Repos
 }
 
-func NewGitGenerator(repos services.Apps) Generator {
+func NewGitGenerator(repos services.Repos) Generator {
 	g := &GitGenerator{
 		repos: repos,
 	}
--- a/pkg/generators/git.go
+++ b/pkg/generators/git.go
@@ -88,29 +88,38 @@ func (g *GitGenerator) generateParamsForGitFiles(appSetGenerator *argoprojiov1al
 	res := []map[string]string{}
 
 	for _, path := range allPaths {
-		content, err := g.repos.GetFileContent(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, path)
+		params, err := g.generateParamsFromGitFile(appSetGenerator, path)
 		if err != nil {
 			return nil, err
-		}
-
-		config := make(map[string]interface{})
-		err = json.Unmarshal(content, &config)
-		if err != nil {
-			return nil, err
-		}
-
-		flat, err := flatten.Flatten(config, "", flatten.DotStyle)
-		if err != nil {
-			return nil, err
-		}
-		params := make(map[string]string)
-		for k, v := range flat {
-			params[k] = v.(string)
 		}
 
 		res = append(res, params)
 	}
 	return res, nil
+}
+
+func (g *GitGenerator) generateParamsFromGitFile(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator, path string) (map[string]string, error) {
+	content, err := g.repos.GetFileContent(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, path)
+	if err != nil {
+		return nil, err
+	}
+
+	config := make(map[string]interface{})
+	err = json.Unmarshal(content, &config)
+	if err != nil {
+		return nil, err
+	}
+
+	flat, err := flatten.Flatten(config, "", flatten.DotStyle)
+	if err != nil {
+		return nil, err
+	}
+	params := make(map[string]string)
+	for k, v := range flat {
+		params[k] = v.(string)
+	}
+
+	return params, nil
 }
 
 func (g *GitGenerator) filterApps(Directories []argoprojiov1alpha1.GitDirectoryGeneratorItem, allApps []string) []string {
--- a/pkg/generators/git_test.go
+++ b/pkg/generators/git_test.go
@@ -29,7 +29,19 @@ func (a argoCDServiceMock) GetApps(ctx context.Context, repoURL string, revision
 	return args.Get(0).([]string), args.Error(1)
 }
 
-func TestGitGenerateParams(t *testing.T) {
+func (a argoCDServiceMock) GetPaths(ctx context.Context, repoURL string, revision string, pattern string) ([]string, error) {
+	args := a.mock.Called(ctx, repoURL, revision, pattern)
+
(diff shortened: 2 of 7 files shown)


dgoodwin:
Looks like the implementation for this PR isn't complete yet but so far looks fine.

Needs a conflict resolved.

rtnpro:
@OmerKahani @dgoodwin One of the design decision I am faced with is if we should implement our own `git` utility for fetching the git repo and scan it's paths for finding the config `json` files.

Initially, I was leaning on using the `argocd` reposerver apiclient to fetch all the paths from the target repository, but it seems that ArgoCD repo server does not allow listing paths, but only `application`s. I am not sure if it will be possible or optimal to try to propose adding an API endpoint in ArgoCD server to list all **paths** in a git repository.

So, for the time being, I am thinking to just hit the ArgoCD reposerver to fetch the creds needed to clone a git repo like here:
https://github.com/argoproj/argo-cd/blob/89ee234634c10312d2fcb74c30e9d590fe01d4ff/server/application/application.go#L1279

and initialize a `git` client to clone the git repo. **This sounds like a duplicate effort as ArgoCD reposerver already has a cache of the git repos, but I see the only way forward, for now, is to implement our little `git` client.**

What do you think?

OmerKahani:
We had the same issue in the git directory genretor. I think we need to add the funcnality to the repo service, but this can take more time.

I did add a [repo_service facade](https://github.com/argoproj-labs/applicationset/blob/master/pkg/services/repo_service.go), you can added the needed funcnality to it, and then we can work on migrating the code to the argocd repo service. This way it will be easy to refactor the code once we have it in the argocd repo service.

@alexec @xianlubird what do you think?

xianlubird:
@OmerKahani agree

rtnpro:
I was away for 2 weeks, as I became a father. I am back now, and have resumed work on this pull request.

OmerKahani:
> I was away for 2 weeks, as I have become a father. I am back now, and have resumed work on this pull request.

Mazal Tov!

pkg/services/repo_service.go
```diff
@@ -19,20 +23,22 @@ type RepositoryDB interface {
 }
 
 type argoCDService struct {
-	repositoriesDB 		RepositoryDB
-	repoClientset 		apiclient.Clientset
+	repositoriesDB RepositoryDB
+	repoClientset  apiclient.Clientset
 }
 
 type Apps interface {
```
rtnpro, line 30: I think we should rename this interface to `Repos` instead of `Apps`, as we are doing more than just fetch `App` data now. We also call the reference of this interfaces instance as `repos`.

pkg/services/repo_service.go
```diff
@@ -66,3 +72,76 @@ func (a *argoCDService) GetApps(ctx context.Context, repoURL string, revision st
 
 	return res, nil
 }
+
+func (a *argoCDService) checkoutRepo(gitClient git.Client, revision string) error {
```
rtnpro, line 76: I am not sure how best to write unit tests for these. We are not doing anything extra on our side, but just call the ArgoCD `git` util, which is unit tested.

rtnpro (commented):
@OmerKahani I got the Git files discovery generator working. I need a review here before I start working on the writing tests. Also, I am a bit confused on what tests to add for `repo_service.go` as I have added almost negligible logic in our code, but using the functions exposed by ArgoCD's `git` util.

May be, we can mock out the git helpers we added: `GetPaths`, `GetFileContent`, and do some functional testing in `git_test.go`.

pkg/services/repo_service.go
OmerKahani: I don't think we should aim to 100% coverage,

why does the function get's a git client and not creating one based on the repo?

pkg/services/repo_service.go
OmerKahani: +1 for that

pkg/services/repo_service.go
```diff
@@ -66,3 +72,76 @@ func (a *argoCDService) GetApps(ctx context.Context, repoURL string, revision st
...
+		return errors.Wrap(err, "Error during fetching commitSHA")
+	}
+	err = gitClient.Checkout(commitSHA)
+	if err != nil {
+		return errors.Wrap(err, "Error during repo checkout")
+	}
+	return nil
+}
+
+func (a *argoCDService) GetPaths(ctx context.Context, repoURL string, revision string, pattern string) ([]string, error) {
```
OmerKahani, line 98: Can we use GetPathes to get the applications? Does get path "*" will return all the top level directories and files?

pkg/generators/git.go
```diff
@@ -36,26 +39,81 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
...
 
-	res := g.generateParams(requestedApps, appSetGenerator)
+	res := g.generateParamsFromApps(requestedApps, appSetGenerator)
 
 	return res, nil
 }
 
-func (g *GitGenerator) filter(Directories []argoprojiov1alpha1.GitDirectoryGeneratorItem, allApps []string) []string {
+func (g *GitGenerator) generateParamsForGitFiles(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error) {
+	allPaths := []string{}
```
OmerKahani, line 79: This function is going to create a new git client for every file. I am not sure if this is a costly code (does it uses network / IO for it?), but if so, we should probably initialize it only once

pkg/generators/git.go
```diff
@@ -36,26 +39,81 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
...
+		flat, err := flatten.Flatten(config, "", flatten.DotStyle)
+		if err != nil {
+			return nil, err
+		}
+		params := make(map[string]string)
+		for k, v := range flat {
+			params[k] = v.(string)
+		}
+
+		res = append(res, params)
```
OmerKahani, line 111: It might be easier, for testing, to extract this to a new function.

pkg/services/repo_service.go
rtnpro: I was hoping to mock the `gitClient` for testing. But, just sending the `repo` object makes more sense.

pkg/services/repo_service.go
rtnpro: `GetPaths` uses `git ls-files` under the hood, and it returns all the paths in the directory tree, matching the specified pattern, and not just the top level files or directories. So, it will need some extra work to retrieve applications using `ls-files`. We will also need to check if a directory is a valid `app` directory, by checking if it contains `helm` or `kustomize` artifacts. Since, ArgoCD repo server already returns the `apps` present in a repo, I think we should just piggy back on that functionality.

pkg/generators/git.go
rtnpro: The `git` checkout call creates a local `tmp` directory, and it gets reused for subsequent calls, until the `tmp` directory is deleted. 

Every `git` `fetch` call will involve network I/O to the git servers, but it will not pull the repo from scratch. So, I think we are good here.

pkg/services/repo_service.go
rtnpro: `gitClient` is actually tied to a `repo` instance, and it's not a generic `git` client like the name suggests. I sent `gitClient` with the hope of mocking it out for testing. But, there's not much to test here anyways, as the `git` client is already tested in ArgoCD codebase. I agree with you that sending `repo` instance makes more sense.

pkg/services/repo_service.go
OmerKahani: An application can be plain k8s objects (without Helm / Kustomize). Currently, the repo server will not return such directories, that's was the reason I asked. We should think about it after you merge this code.

pkg/services/repo_service.go
rtnpro: @OmerKahani I gave it a lot of thought to pass `repo` object to `checkoutRepo` func, but there is a use case where `gitClient` is used later, e.g., https://github.com/argoproj-labs/applicationset/pull/45/files#diff-004c4dc22d5fb77bff9d0d6fd29d999d0e068fc99b996bf91a75213585a040efR93. It did not make sense to initialize gitClient multiple times.

So, what I did here: https://github.com/argoproj-labs/applicationset/pull/45/commits/e4e1e5d57f132790d9b9faef2fd8ae510a21cd14 is rename `gitClient` to `getRepoClient` which explicitly states the role of the variable. The git client instance is tied to a repo.

pkg/generators/git_test.go
```diff
@@ -74,26 +95,101 @@ func TestGitGenerateParams(t *testing.T) {
...
+			expectedError: nil,
+		},
 	}
 
 	for _, c := range cases {
 		cc := c
 		t.Run(cc.name, func(t *testing.T) {
 			argoCDServiceMock := argoCDServiceMock{}
-			argoCDServiceMock.On("GetApps", mock.Anything, mock.Anything, mock.Anything).Return(c.repoApps, c.repoError)
+			mockCalls := argoCDServiceMock.
```
rtnpro, line 184: @OmerKahani I started writing test cases for git file generator scenario, however I am not able to use `testify.Mock` properly. What I am trying to do is setup multiple expectations per test case:

- One for `GetPaths`
- Many for `GetFileContent`

I tried to use Mock chaining as mentioned here: https://godoc.org/github.com/stretchr/testify/mock#Call.On, but still could not get it working. The test is failing with error below:

```
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
        A future release of golang/protobuf will delete this package,
        which has long been excluded from the compatibility promise.

time="2020-11-18T21:58:55+05:30" level=info msg="applications result from the repo service" allAps="[app1 app2 p1/app3]" repoURL=RepoURL revision=Revision total=3
time="2020-11-18T21:58:55+05:30" level=info msg="applications result from the repo service" allAps="[app1 p1/app2 p1/p2/app3 p1/p2/p3/app4]" repoURL=RepoURL revision=Revision total=4
time="2020-11-18T21:58:55+05:30" level=info msg="applications result from the repo service" allAps="[]" repoURL=RepoURL revision=Revision total=0
--- FAIL: TestGitGenerateParams (0.00s)
    --- FAIL: TestGitGenerateParams/happy_flow_-_created_apps (0.00s)
        git_test.go:220: PASS:  GetApps(string,string,string)
        git_test.go:220: FAIL:  GetPaths(string,string,string,string)
                        at: [git_test.go:186]
        git_test.go:220: FAIL: 1 out of 2 expectation(s) were met.
                The code you are testing needs to make 1 more call(s).
                at: [git_test.go:220]
    --- FAIL: TestGitGenerateParams/It_filters_application_according_to_the_paths (0.00s)
        git_test.go:220: PASS:  GetApps(string,string,string)
        git_test.go:220: FAIL:  GetPaths(string,string,string,string)
                        at: [git_test.go:186]
        git_test.go:220: FAIL: 1 out of 2 expectation(s) were met.
                The code you are testing needs to make 1 more call(s).
                at: [git_test.go:220]
    --- FAIL: TestGitGenerateParams/handles_empty_response_from_repo_server (0.00s)
        git_test.go:220: PASS:  GetApps(string,string,string)
        git_test.go:220: FAIL:  GetPaths(string,string,string,string)
                        at: [git_test.go:186]
        git_test.go:220: FAIL: 1 out of 2 expectation(s) were met.
                The code you are testing needs to make 1 more call(s).
                at: [git_test.go:220]
    --- FAIL: TestGitGenerateParams/handles_error_from_repo_server (0.00s)
        git_test.go:220: PASS:  GetApps(string,string,string)
        git_test.go:220: FAIL:  GetPaths(string,string,string,string)
                        at: [git_test.go:186]
        git_test.go:220: FAIL: 1 out of 2 expectation(s) were met.
                The code you are testing needs to make 1 more call(s).
                at: [git_test.go:220]
    --- FAIL: TestGitGenerateParams/handles_error_from_repo_server#01 (0.00s)
        git_test.go:220: PASS:  GetApps(string,string,string)
        git_test.go:220: FAIL:  GetPaths(string,string,string,string)
                        at: [git_test.go:186]
        git_test.go:220: FAIL: 1 out of 2 expectation(s) were met.
                The code you are testing needs to make 1 more call(s).
                at: [git_test.go:220]
FAIL
FAIL    command-line-arguments  1.222s
FAIL

```

Removing or commenting out the new `ArgoCDServiceMock.On` calls allows the test to pass.

rtnpro: @dgoodwin I need some help here.

pkg/services/repo_service.go
OmerKahani: OK, sounds good

pkg/generators/git_test.go
rtnpro: @OmerKahani ^^

OmerKahani: Hi, sorry for the delay it was a busy week.

The problem is with line 186: https://github.com/argoproj-labs/applicationset/pull/45/files/1f6c6234c6514ba4d0fbf5693699c3c5ece966cc#diff-c1129e2ddfe8072c3267b62d8ac75bcd5dfa92143c40c478527853c08752b7eeR186

I think it will be best to create a new TestGitGenerateParams so there will be two functions:
TestGitGenerateParamsFromDirectory
TestGitGenerateParamsFromFile

everyone will have its own test cases and mock only one of the functions in the argocd service

Does it sound reasonable to you?

pkg/services/repo_service.go
```diff
@@ -66,3 +72,75 @@ func (a *argoCDService) GetApps(ctx context.Context, repoURL string, revision st
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = checkoutRepo(gitRepoClient, revision)
+	if err != nil {
+		return nil, err
+	}
+
+	paths, err := gitRepoClient.LsFiles(pattern)
```
jgwest, line 93: The `LsFiles` implementation from Argo CD's git client does not handle recursive filters (eg`**/*.json` to recursively match all JSON files in a repo)

`LsFiles` is a wrapper around `git ls-files --full-name -- (filter)`, which treats `**/*.json` as equivalent to `*/*.json` (only matching from root folder)

pkg/generators/git.go
```diff
@@ -36,26 +39,90 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
...
+}
+
+func (g *GitGenerator) generateParamsFromGitFile(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator, path string) (map[string]string, error) {
+	content, err := g.repos.GetFileContent(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, path)
+	if err != nil {
+		return nil, err
+	}
+
+	config := make(map[string]interface{})
+	err = json.Unmarshal(content, &config)
```
jgwest, line 108: It looks like this only handles JSON data from the git repo, whereas the proposal notes that "referenced files would contain a json/yaml list of arbitrary structured objects"."

pkg/generators/git.go
```diff
@@ -36,26 +39,90 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
...
 
-func (g *GitGenerator) filter(Directories []argoprojiov1alpha1.GitDirectoryGeneratorItem, allApps []string) []string {
+func (g *GitGenerator) generateParamsForGitFiles(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) ([]map[string]string, error) {
+	allPaths := []string{}
+	for _, requestedPath := range appSetGenerator.Git.Files {
+		paths, err := g.repos.GetPaths(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, requestedPath.Path)
+		if err != nil {
+			return nil, err
+		}
+		allPaths = append(allPaths, paths...)
```
jgwest, line 85: I'm guessing we may want to remove any duplicate files from the list, in the case where multiple `path` patterns are specified and they happen to (partially) overlap between one another.

jgwest (changes requested):

pkg/services/repo_service.go
jgwest: And related: realized that since Argo CD's `util/git` utility calls to `git` directly, `git` needs to be included in the [Dockerfile](https://github.com/argoproj-labs/applicationset/blob/master/Dockerfile)

pkg/generators/git_test.go
rtnpro: Yes. Created separate test function for testing Git file generator.

pkg/services/repo_service.go
rtnpro: @jgwest I ran `git ls-files --full-name -- **/*.json` on https://github.com/rtnpro/argocd-applicationset-examples/ and it returned

```
cluster-config/engineering/local/config.json
```

So, it supports searching for matching files in nested directories.

rtnpro: @jgwest I have added more `.json` files to the repo: https://github.com/rtnpro/argocd-applicationset-examples/ to validate the output of `git ls-files --full-name -- <pattern>` and it seems to work as expected.

image

cc @OmerKahani

pkg/generators/git.go
rtnpro: @jgwest Fixed by https://github.com/argoproj-labs/applicationset/pull/45/commits/69a891ae4a8b4ab6dd133834c22b40feb0eb6fc7

rtnpro:
@OmerKahani @jgwest 

Apart from agreeing on whether to use `git ls-files ...` to find matching paths in a git repo, I have addressed all other review requests. I am confident that we are ready with this pull request.

Regarding the shortcomings mentioned by @jgwest in using `git ls-files ...` to list file paths matching a pattern, I tried reproducing the issues reported, but could not do so. Ref: https://github.com/argoproj-labs/applicationset/pull/45#discussion_r547343762

I am using git version `2.28.0`.

pkg/services/repo_service.go
jgwest: @rtnpro Agreed, thanks for investigating, it appears to work as expected for me as well... I'm not able to reproduce with the information I originally provided, and I'm not able to reproduce now (and no longer have the environment I originally used),  so this can be resolved.

pkg/generators/git.go
jgwest: :+1:

jgwest (approved):
Thanks @rtnpro, LGTM! Any outstanding items that arise can be handled under separate issues/PRs as needed.

OmerKahani (approved):
@rtnpro sorry for all of the delays. This was a great effort from your side and a huge contribution to the project.

xianlubird (approved):

PR #48, opened 2020-09-17
add sync status

OmerKahani:
Add the option to determine the sync status. When using the git generator, a  user can cause an application to be deleted by changing the name of the folder. This PR aimed to give an option to prevent delete / update in the reconcile. 

It's should probably be part of the generator metadata, but for now, we can start with a global flag.

Based on the policies from external-dns

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"github.com/argoproj-labs/applicationset/pkg/generators"
 	"github.com/argoproj-labs/applicationset/pkg/services"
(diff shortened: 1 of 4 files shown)


xianlubird:
pls fix the conflict

dgoodwin:
Are we confident exposing this as a global setting on the CLI isn't going to get picked up by someone and used beyond what we'd like? This really should be in the generator spec as you mentioned. Doing it across the whole controller is definitely not something we'd want to live for very long.

xianlubird (approved):

Issue #52, opened 2020-09-25
Reconciler error - ApplicationSet is empty


jbaskeen:
I am running the ArgoCD v1.6.1+159674e via the Operator in OpenShift 4.5.4. When using one of the ApplicationSet examples I get the following error in the argocd-applicationset-controller and the Applications are not created.

I have REDACTED some server name info from the log and the configs.

Log

```
2020-09-24T11:10:49.333Z	INFO	setup	using argocd namespace	{"namespace": "argo"}
2020-09-24T11:10:52.940Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8081"}
2020-09-24T11:10:52.943Z	INFO	setup	starting manager
2020-09-24T11:10:52.943Z	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}
2020-09-24T11:10:53.044Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "applicationset", "source": "kind source: /, Kind="}
2020-09-24T11:10:53.145Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "applicationset", "source": "kind source: /, Kind="}
2020-09-24T11:10:53.145Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "applicationset", "source": "kind source: /, Kind="}
2020-09-24T11:10:53.245Z	INFO	controller-runtime.controller	Starting Controller	{"controller": "applicationset"}
2020-09-24T11:10:53.245Z	INFO	controller-runtime.controller	Starting workers	{"controller": "applicationset", "worker count": 1}
time="2020-09-24T11:10:53Z" level=info msg="processing event for cluster secret" name=REDACTED namespace=argo type=createSecretEventHandler
time="2020-09-24T11:10:53Z" level=info msg="listed ApplicationSets" count=1 type=createSecretEventHandler
time="2020-09-24T11:10:53Z" level=info msg="processing event for cluster secret" name=REDACTED namespace=argo type=createSecretEventHandler
time="2020-09-24T11:10:53Z" level=info msg="listed ApplicationSets" count=1 type=createSecretEventHandler
time="2020-09-24T11:10:53Z" level=info msg="generated 2 applications" generator="&{}"
time="2020-09-24T11:10:53Z" level=info msg="generated 2 applications" generator="&{0xc0003a5320}"
[...]

PR #53, opened 2020-09-25
Drop cluster-wide RBAC install and fix finalizer permissions issue.

dgoodwin:
We picked up an ArgoCD pattern where you can install with namespaced or
cluster-wide permissions. The cluster-wide was required so Argo could
potentially sync resources to the cluster it's running in. This however
is not relevant to the ApplicationSet CR, so we do not need to support
dual modes of install and can instead just deploy with namespaced RBAC.

Fix a missing permission for finalizers so we can set an Application
blockOwnerDeletion reference back to the ApplicationSet that is
originated from.


dgoodwin:
/cc @mgoodness  

As we discussed dropping this cluster scoped install.

dgoodwin:
Fixes this error which surfaces at least with cluster generator:

```
time="2020-09-25T14:37:30Z" level=info msg="matched cluster secret" cluster=cluster-api.dgood-dev.new-installer.openshift.com-3927666468
time="2020-09-25T14:37:30Z" level=info msg="generated 1 applications" generator="&{0xc0003c5980}"
time="2020-09-25T14:37:30Z" level=error msg="failed to unchanged Application" action=unchanged app=admin-configmaps appSet=clustergen error="applications.argoproj.io \"admin-configmaps\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"
2020-09-25T14:37:30.114Z        ERROR   controller-runtime.controller   Reconciler error        {"controller": "applicationset", "name": "clustergen", "namespace": "argocd", "error": "applications.argoproj.io \"admin-configmaps\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
```

OmerKahani:
@dgoodwin does it fix #52 ?

jbaskeen:
@dgoodwin @OmerKahani the fix here to add `applicationsets/finalizers`  in`base/rbac` resolved my final issue #52

dgoodwin:
Fixes #52

xianlubird (approved):

Issue #32, opened 2020-09-02
Add SyncPolicy InitalSync


OmerKahani:
From the Proposal https://docs.google.com/document/d/1juWGr20FQaJmuuTIS8mBFmWWDU422M_FQMuhp5c1jt4/edit#

dgoodwin:
Does SyncPolicy.InitialSync have any relation to ApplicationSets? In ArgoCD you get those Applications in the UI that you can manually sync, but that's not really a thing here with ApplicationSet. We'd just log what we generated and you'd get no result, which doesn't seem overly useful.

OmerKahani:
creation alone doesn't really do anything (no pods...), so I can see how initial sync makes sense. But maybe we should wait to see how https://github.com/argoproj/argo-cd/issues/1283 is resolved before implementing this

dgoodwin:
@OmerKahani can you clarify for me, what would SyncPolicy.InitialSync do on ApplicationSet?

Application.Spec.SyncPolicy can be used by embedding it within your ApplicationSet.Spec.Template.

But what would an ApplicationSet.Spec.SyncPolicy.InitialSync do? We don't have the concept of manual syncs for an ApplicationSet to it's Applications.

OmerKahani:
Adding an application be iteslf doesn't give any bussins value. In case we added a cluster and a cluster secret, our cluster genretor will create the applications for the cluster, but then we still need to go into argocd and click sync on every application.

InistialSync complate the bussins value cycle, so by adding a cluster secret you will both the applications and the real K8s objects.

dgoodwin:
This can be accomplished by using ApllicationSet.Spec.Template.SyncPolicy.InitialSync though, we embed an Application template which supports initial sync, and that is handled for us automatically.

This is why I don't think we need a separate field for ApplicationSet.Spec.SyncPolicy.InitialSync, I think including this in the design document may have been a copy-paste mistake.

(note "Template" is in one of the examples above, and not in the other)

dgoodwin:
cc @jessesuen to check if the design doc intentionally included a separate InitialSync param on ApplicationSet vs just using the one we get for free on ApplicationSet.Spec.Template.

wnagele:
@dgoodwin where do you see this option? I cannot find anything on the [template](https://github.com/argoproj-labs/applicationset/blob/85f8d904cadb9595656dd12c312310ef00a27099/api/v1alpha1/applicationset_types.go#L50).

[...]

PR #55, opened 2020-09-25
Refine ApplicationSet SyncPolicy API to be less like ArgoCD.

dgoodwin:
The ArgoCD core Application.Spec.SyncPolicy is available as part of the
embedded ApplicationSet.Spec.Template.

However we inherited some of this API within the ApplicationSet.Spec
where it does not make sense.

This PR drops the "automated" sub-struct, as automated is really the
only sync policy for an ApplicationSet. We do not support the concept of
a manual Sync from a UI or CLI tool as we would in ArgoCD itself, and as
far as I know this is not desirable.

This PR also drops the "initialSync" field, as I do not believe this
concept makes sense for an ApplicationSet. Applications are always
synced for their ApplicationSet.

ApplicationSet.Spec.SyncPolicy.Automated.Prune then moves up to
ApplicationSet.Spec.SyncPolicy.Prune.


dgoodwin:
This PR is for consideration and may be controversial, or I may be missing vision on how this is intended to be used. Core assumption I'm making here is that there will be no UI around ApplicationSets where you click a button to reconcile to your Applications. It just happens via the standard watch/reconcile controller loop. 

If I am mistaken about this then this PR should be closed.

However if not I think the API should be adjusted to better reflect what is possible and prevent Application SyncPolicy concepts from slipping in.

Additionally I changed Prune to SkipPrune, as I believe pruning should be the default behavior, and thus the go default of false would better align with SkipPrune.

manifests/crds/argoproj.io_applicationsets.yaml
```diff
@@ -1,13 +1,11 @@
+
```
dgoodwin, line 1: Renamed as this is where "make generate" is dropping the file, not sure why the mismatch.

OmerKahani:
Related to #32

manifests/crds/argoproj.io_applicationsets.yaml
```diff
@@ -1,13 +1,11 @@
...
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
-  labels:
```
xianlubird, line 8: why delete this

dgoodwin: This is the autogenerated output from make generate, typically projects would let these be generated rather than maintained manually.

However was this file being maintained manually? It might explain the filename that confused me.

Or were these labels injected and I'm using an incorrect version of something.

xianlubird (approved):

PR #56, opened 2020-09-26
Add template override

### LINKED ISSUE #27: Implement Template Override
Implement this example https://github.com/argoproj-labs/applicationset/blob/master/examples/template-override.yaml

OmerKahani:
fix #27


pkg/generators/interface.go
```diff
@@ -18,6 +18,8 @@ type Generator interface {
 	// In case there is more then one generator the time will be the minimum of the times.
 	// In case NoRequeueAfter is empty, it will be ignored
 	GetRequeueAfter(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) time.Duration
+
+	GetTemplate(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) *argoprojiov1alpha1.ApplicationSetTemplate
```
jgwest, line 22: Suggest adding a  comment got `GetTemplate` here, something like `// GetTemplate returns the 'Template' field within the generator's CRD element`

examples/template-override.yaml
```diff
@@ -8,33 +8,41 @@ metadata:
   name: guestbook
 spec:
   generators:
-  - git:
-      repoURL: https://github.com/infra-team/cluster-deployments.git
-      files:
-      - path: "**/kustomization.yaml"
+  - list:
```
jgwest, line 11: It looks like you're replacing the Git files generator example with a list generator example, but I assume the original example is still valid? Another option could be a separate `template-overrides-list.yaml` that contains the list changes.

pkg/controllers/applicationset_controller.go
```diff
@@ -265,11 +267,17 @@ func getTempApplication(applicationSetTemplate argoprojiov1alpha1.ApplicationSet
 	return &tmplApplication
 }
 
+func mergeGeneratorTemplate(g generators.Generator, requestedGenerator *argoprojiov1alpha1.ApplicationSetGenerator, applicationSetTemplate argoprojiov1alpha1.ApplicationSetTemplate) argoprojiov1alpha1.ApplicationSetTemplate {
+	dest := g.GetTemplate(requestedGenerator)
```
jgwest, line 271: How does this change look:
```go
func mergeGeneratorTemplate(g generators.Generator, requestedGenerator *argoprojiov1alpha1.ApplicationSetGenerator, applicationSetTemplate argoprojiov1alpha1.ApplicationSetTemplate) (argoprojiov1alpha1.ApplicationSetTemplate, error) {

	// Make a copy of the value from `GetTemplate()` before merge, rather than copying directly into
	// the provided parameter (which will touch the original resource object returned by client-go)
	dest := g.GetTemplate(requestedGenerator).DeepCopy()

	err := mergo.Merge(dest, applicationSetTemplate)

	return *dest, err
}
```

This: 
- Makes a deep copy of the `requestedGenerator` parameter before merging, so we don't accidentally modify it with the `Merge` call
- Return the error from `Merge`

Sound good?

jgwest (approved):
Thanks @OmerKahani, looks great, just a couple suggestions, let me know what you think.

pkg/controllers/applicationset_controller.go
OmerKahani: thanks

examples/template-override.yaml
OmerKahani: The Git example is harder to test locally that's why I changed it. I'm also working on a git example that will work. 
Do you think we can merge it in the meantime?

jgwest: Sure, sounds good.

jgwest (approved):
LGTM!

PR #58, opened 2020-10-01
Annotations

### LINKED ISSUE #31: Add support to metadata.annotations in the template
From the Proposal https://docs.google.com/document/d/1juWGr20FQaJmuuTIS8mBFmWWDU422M_FQMuhp5c1jt4/edit#

mgoodness:
- Adds annotations from template to rendered Applications
- Makes cluster annotations available for templating (Closes #31)

### DIFF
--- a/pkg/generators/cluster.go
+++ b/pkg/generators/cluster.go
@@ -26,7 +26,6 @@ type ClusterGenerator struct {
 	client.Client
 }
 
-
 func NewClusterGenerator(c client.Client) Generator {
 	g := &ClusterGenerator{
 		Client: c,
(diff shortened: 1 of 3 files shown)


dgoodwin:
Would it be feasible to add a small unit test just to ensure we keep it working? It looks relatively straightforward to add in to cluster_test.go. If time is an issue we can merge and add later.

pkg/generators/cluster.go
```diff
@@ -65,9 +64,12 @@ func (g *ClusterGenerator) GenerateParams(
 
 	res := make([]map[string]string, len(clusterSecretList.Items))
 	for i, cluster := range clusterSecretList.Items {
-		params := make(map[string]string, len(cluster.ObjectMeta.Labels) + 2)
+		params := make(map[string]string, len(cluster.ObjectMeta.Labels)+2)
```
OmerKahani, line 67: the length of the map should be: len(cluster.ObjectMeta.Labels) + len(cluster.ObjectMeta.Annotations) + 2

OmerKahani (changes requested):

pkg/generators/cluster.go
mgoodness: Fixed. Thanks!

OmerKahani (approved):

xianlubird (approved):

PR #64, opened 2020-11-24
Add build and test GH action, fix TestGetApps race condition (#83)

jgwest:
This PR:
- Adds a GitHub action to the workflow that runs the build (both docker and standalone), and then runs `go test` (the job/steps  are [based on Argo CD's](https://github.com/argoproj/argo-cd/blob/master/.github/workflows/ci-build.yaml))
- Fixes the `TestGetApps` test, which was intermittently failing due to string ordering

Failing test:
```
2020/11/24 12:08:52 proto: tag has too few fields: "-"
--- FAIL: TestGetApps (0.00s)
    --- FAIL: TestGetApps/Happy_Flow (0.00s)
        repo_service_test.go:156: 
            	Error Trace:	repo_service_test.go:156
            	Error:      	Not equal: 
            	            	expected: []string{"app2", "app1"}
            	            	actual  : []string{"app1", "app2"}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,4 +1,4 @@
            	            	 ([]string) (len=2) {
            	            	- (string) (len=4) "app2",
            	            	- (string) (len=4) "app1"
            	            	+ (string) (len=4) "app1",
            	            	+ (string) (len=4) "app2"
            	            	 }
            	Test:       	TestGetApps/Happy_Flow
FAIL
```

PR #67, opened 2020-12-01
Log warning if applicationset contains unrecognized generators

jopit:


pkg/controllers/applicationset_controller.go
```diff
@@ -67,6 +70,23 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	// Log a warning if there are unrecognized generators
```
jgwest, line 73: Nitpick: I think this block can be moved into its own function (eg `checkInvalidGenerators`), which can then be called from here.

pkg/controllers/applicationset_controller.go
```diff
@@ -67,6 +70,23 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
+	hasInvalidGenerators, invalidGenerators := invalidateGenerators(&applicationSetInfo)
+	if len(invalidGenerators) > 0 {
+		gnames := make([]string, 0, len(invalidGenerators))
```
jgwest, line 76: Nitpick: this is perfectly valid but `gnames := []string{}` is a bit cleaner

pkg/controllers/applicationset_controller.go
```diff
@@ -100,6 +120,68 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
+	generator, ok := generators[index].(map[string]interface{})
+	if !ok {
+		log.Warnf("coundn't get generator from kubectl.kubernetes.io/last-applied-configuration annotation")
+		return
+	}
+
+	var name string
+	for key := range generator {
+		name = key
+		names[name] = true
```
jgwest, line 180: Nitpick: think this can just be `names[key] = true`

jgwest (changes requested):
Thanks @jopit, code LGTM, just some nitpicks I've mentioned below... It would also be good to add a unit test for this case, for example testing the `invalidateGenerators` function. (Once https://github.com/argoproj-labs/applicationset/pull/64/files merges then tests runs will be included in the PR build with a `make test`, but in the mean time we have to do `go test ./...`)

jopit:
@jgwest thanks for the review.  I've made changes based on your feedback.

pkg/controllers/applicationset_controller.go
jopit: Done

pkg/controllers/applicationset_controller.go
jopit: Done

pkg/controllers/applicationset_controller.go
jopit: Done

jgwest (approved):
Thanks @jopit, LGTM and approved!

xianlubird (approved):

PR #69, opened 2020-12-09
Raise error for duplicate application names

### LINKED ISSUE #68: ApplicationSet which generates Applications with matching names should report an error and return
As of this writing, the `generateApplications` method does not check for applications with duplicate names, and thus multiple generated Applications may use the same name.

As discussed in the comments of the design proposal, this should be reported as an error:
> Two application with the same name would be considered an error. In this case error would be saved in status as an error condition.

### To reproduce:

Apply the following YAML, and observe that it reports `generated 3 applications`, even though only 1 Application resource is created:
```yaml
# The list generator specifies a literal list of argument values to the app spec template.
kind: ApplicationSet
metadata:
  name: guestbook
spec:
  generators:
  - list:
      elements:
      - cluster: engineering-dev
        url: https://1.2.3.4
      - cluster: engineering-prod
        url: https://2.4.6.8
      - cluster: finance-preprod
        url: https://9.8.7.6
  template:
    metadata:
      name: 'matching-guestbook'
    spec:
      project: default
      source:
        repoURL: https://github.com/jgwest/argocd-example-apps.git
        targetRevision: HEAD
        path: guestbook
      destination:
        namespace: guestbook
```

In this case, 0 applications should be generated, and an error should be reported in the ApplicationSet status conditions.

jopit:
Signed-off-by: John Pitman <jpitman@redhat.com>

Fixes #68

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -77,6 +78,10 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 	desiredApplications, err := r.generateApplications(applicationSetInfo)
 	if err != nil {
 		return ctrl.Result{}, err
+	}
+	err = validateDesiredApplications(desiredApplications)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("Reconciling ApplicationSet %s: %w", applicationSetInfo.Name, err)
 	}
 
 	if r.Policy.Update() {
@@ -188,6 +193,17 @@ func addInvalidGeneratorNames(names map[string]bool, applicationSetInfo *argopro
 		names[key] = true
 		break
 	}
+}
+
+func validateDesiredApplications(desiredApplications []argov1alpha1.Application) error {
+	for i1, app1 := range desiredApplications {
+		for i2, app2 := range desiredApplications {
+			if i1 != i2 && app1.Name == app2.Name {
+				return errors.New("application set contains applications with duplicate names")
+			}
+		}
+	}
+	return nil
 }
 
 func (r *ApplicationSetReconciler) GetRelevantGenerators(requestedGenerator *argoprojiov1alpha1.ApplicationSetGenerator) []generators.Generator {
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -206,12 +206,12 @@ func addInvalidGeneratorNames(names map[string]bool, applicationSetInfo *argopro
 }
 
 func hasDuplicateNames(applications []argov1alpha1.Application) (bool, string) {
-	for i1, app1 := range applications {
-		for i2, app2 := range applications {
-			if i1 != i2 && app1.Name == app2.Name {
-				return true, app1.Name
-			}
+	nameSet := map[string]struct{}{}
+	for _, app := range applications {
+		if _, present := nameSet[app.Name]; present {
+			return true, app.Name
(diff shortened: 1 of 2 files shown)


jgwest (changes requested):
Thanks @jopit, looks great, just a couple quick tweaks.

pkg/controllers/applicationset_controller.go
```diff
@@ -78,6 +79,10 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	err = validateDesiredApplications(desiredApplications)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("Reconciling ApplicationSet %s: %w", applicationSetInfo.Name, err)
```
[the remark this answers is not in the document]
jgwest: (New issue: https://github.com/argoproj-labs/applicationset/issues/71)

pkg/controllers/applicationset_controller.go
```diff
@@ -78,6 +79,10 @@ func (r *ApplicationSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	err = validateDesiredApplications(desiredApplications)
```
OmerKahani, line 82: I think this can be part of the `generateApplications`, if there is a duplicate name the generate will fail with an error. what do you think?

jopit: I think it's cleaner if the validation is done separately from the generation, particularly in light of Jonathan's comments above about the reconcile being retried when there is an error.  If validation is separate, an error from `generateApplications` will still result in the reconcile being retried, while a duplicate application name will not.

Also, if there is a problem generating one of the applications, `generateApplications` currently will return both an error and a slice containing the valid applications (I'm not sure why).  I think weaving the return of an error for duplicate names into this logic would be unnecessarily complicated.

pkg/controllers/applicationset_controller.go
```diff
@@ -190,6 +195,17 @@ func addInvalidGeneratorNames(names map[string]bool, applicationSetInfo *argopro
 	}
 }
 
+func validateDesiredApplications(desiredApplications []argov1alpha1.Application) error {
+	for i1, app1 := range desiredApplications {
+		for i2, app2 := range desiredApplications {
+			if i1 != i2 && app1.Name == app2.Name {
+				return errors.New("application set contains applications with duplicate names")
```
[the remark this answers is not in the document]
jopit: done

pkg/controllers/applicationset_controller.go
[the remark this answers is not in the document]
jopit: done

jgwest (approved):
Thanks @jopit , approved!

pkg/controllers/applicationset_controller.go
```diff
@@ -190,6 +205,17 @@ func addInvalidGeneratorNames(names map[string]bool, applicationSetInfo *argopro
...
 }
 
+func hasDuplicateNames(applications []argov1alpha1.Application) (bool, string) {
+	for i1, app1 := range applications {
+		for i2, app2 := range applications {
+			if i1 != i2 && app1.Name == app2.Name {
+				return true, app1.Name
+			}
+		}
+	}
```
wtam2018, line 215:
```suggestion
	nameSet := map[string]struct{}{}
	for _, app := range applications {
		if _, present := nameSet[app.Name]; present {
			return true, app.Name
		}
		nameSet[app.Name] = struct{}{}
	}
```

wtam2018: We can probably make it O(n) instead of O(n^2).

wtam2018:
LGTM

xianlubird (approved):

PR #89, opened 2021-01-10
Sync go.(mod/sum) with argo-cd 1.8 release

kvendingoldo:


kvendingoldo:
@jgwest @OmerKahani @wtam2018 @xianlubird can you check what is wrong with the unit test for generators/cluster, please? I see that it's the wrong order but in the general result the same.

kvendingoldo:
@jgwest @dgoodwin @mgoodness @OmerKahani @wtam2018 @xianlubird please, review

pkg/utils/createOrUpdate.go
```diff
@@ -11,17 +11,25 @@ import (
...
 
 // override "sigs.k8s.io/controller-runtime" CreateOrUpdate function to add equity function for argov1alpha1.ApplicationDestination
 // argov1alpha1.ApplicationDestination has a private variable, so the default implementation fails to compare it
-func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f controllerutil.MutateFn) (controllerutil.OperationResult, error) {
-	key, err := client.ObjectKeyFromObject(obj)
-	if err != nil {
-		return controllerutil.OperationResultNone, err
+func CreateOrUpdate(ctx context.Context, c client.Client, obj client.Object, f controllerutil.MutateFn) (controllerutil.OperationResult, error) {
+	key := client.ObjectKeyFromObject(obj)
+	if err := c.Get(ctx, key, obj); err != nil {
```
wtam2018, line 22: Did you copy this if-block by mistake?  It looks like an exact copy of the one below it.

jgwest: I had the same question as @wtam2018 , the first and second if block appear to be the same, unless there is a subtle difference hiding in there...

jgwest (changes requested):
Thanks @kvendingoldo, looks great! Check out our question re: `createOrUpdate.go`, and let us know if this was intentional...

pkg/utils/createOrUpdate.go
kvendingoldo: It's copy-paste issue

kvendingoldo:
@wtam2018 @jgwest it's fixed, thank you for your attention :)

jgwest (approved):
LGTM!

PR #91, opened 2021-01-11
feat(cluster-generator): support arbitrary key:value pairs

mgoodness:
Per the cluster generator filters example, we should allow [arbitrary key:value pairs](https://github.com/argoproj-labs/applicationset/blob/master/examples/filters.yaml#L17) that can be [used for templating](https://github.com/argoproj-labs/applicationset/blob/master/examples/filters.yaml#L40).

Depends on #93 


mgoodness:
Premature PR, as the CRD will have to be regenerated with the addition to `ClusterGenerator`. That's complicated by the fact that the project was bootstrapped using kubebuilder v1. I'm in the process of migrating to v2 - a separate PR.

mgoodness:
As reported in Slack, something in the e2e tests doesn't like this new ApplicationSet CRD generated using `apiextensions.k8s.io/v1`. If I generate the CRD using `apiextensions.k8s.io/v1beta` (as in #97), everything is fine.

Can confirm that the `apiextensions.k8s.io/v1` CRD works fine in my cluster.

mgoodness:
Punting on the `apiextensions.k8s.io/v1` for now, in the interest of getting this feature merged.

api/v1alpha1/applicationset_types.go
```diff
@@ -75,6 +75,7 @@ type ClusterGenerator struct {
 	// Clusters today are stored as Kubernetes Secrets, thus the Secret labels will be used
 	// for matching the selector.
 	Selector metav1.LabelSelector `json:"selector,omitempty"`
+	Values   map[string]string    `json:"values,omitempty"`
```
jgwest, line 78: Only a minor request: could you add a comment for the `Values` line, something like:
```
// Values contains key/value pairs which are passed directly as parameters to the template 
```

jgwest (changes requested):
Thanks @mgoodness,  looks great, just one small tweak and good to go.

pkg/generators/cluster.go
```diff
@@ -65,7 +65,10 @@ func (g *ClusterGenerator) GenerateParams(
 	// For each matching cluster secret
 	res := make([]map[string]string, len(clusterSecretList.Items))
 	for i, cluster := range clusterSecretList.Items {
-		params := make(map[string]string, len(cluster.ObjectMeta.Annotations)+len(cluster.ObjectMeta.Labels)+2)
+		params := make(map[string]string, len(appSetGenerator.Clusters.Values)+len(cluster.ObjectMeta.Annotations)+len(cluster.ObjectMeta.Labels)+2)
+		for key, value := range appSetGenerator.Clusters.Values {
+			params[key] = value
```
ygelfand, line 70: any thoughts on prefixing w/ something like
 ```go
        params[fmt.Sprintf("metadata.values.%s", key)] = value
```
 here instead, and avoiding stomping on  other things in params[]?

jbaskeen: +1 for allowing values to be passed through params, many use cases there for templating those params in helm as control variables in a values.yaml

mgoodness: > any thoughts on prefixing w/ something like
> 
> ```go
>        params[fmt.Sprintf("metadata.values.%s", key)] = value
> ```
> 
> here instead, and avoiding stomping on other things in params[]?

I have no objections to that, though I think we should drop the `metadata.` prefix. Could cause confusion with `metadata.labels` and `metadata.annotations` coming from one place in the AppSet and `metadata.value` from another.

I propose:

 ```go
        params[fmt.Sprintf("values.%s", key)] = value
 ```

api/v1alpha1/applicationset_types.go
mgoodness: No problem.

jgwest (approved):
Thanks @mgoodness, still needs a quick rebase but LGTM!

mgoodness:
> Thanks @mgoodness, still needs a quick rebase but LGTM!

Go go go, before someone sneaks in another change!

jgwest:
Merged, thanks again... sorry for the code churn, going forward we should have fewer conflicts now that we are close to having things nailed down in the generated manifests. :smile:

PR #93, opened 2021-01-12
kubebuilder 2

mgoodness:
This is primarily a re-bootstrap of the project using kubebuilder 2, pulling in changes that have been made upstream since June. I also made opportunistic fixes - mostly to address linting errors. The biggest change is that the ApplicationSet CRD has been re-generated to support the latest version of the Argo CD Application API. It also presents an opportunity to move to `github.com/go-logr/logr` from `github.com/sirupsen/logrus` since that's what controller-runtime now uses.

Depends on #96


OmerKahani (approved):

Dockerfile
```diff
@@ -1,17 +1,27 @@
 # Build the binary
-FROM golang:1.14.1 as builder
+FROM golang:1.14.12 as builder
 
 WORKDIR /workspace
+
+# Copy the Go Modules manifests
+COPY go.mod go.mod
```
wtam2018, line 7:
```suggestion
COPY go.mod .
```

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
+COPY go.sum go.sum
```
wtam2018, line 8:
```suggestion
COPY go.sum .
```

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+
 # Copy the go source
-COPY . .
+COPY main.go main.go
+COPY api/ api/
+COPY pkg/ pkg/
```
wtam2018, line 16:
```suggestion
COPY main.go .
COPY api/ .
COPY pkg/ .
```

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-w -s" -a -o applicationset-controller main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -ldflags="-w -s" -a -o applicationset-controller main.go
```
wtam2018, line 19: Do we really need `GO111MODULE=on`?

mgoodness: Good call. Removing.

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
+COPY go.sum .
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+
 # Copy the go source
-COPY . .
+COPY main.go .
+COPY api/ api/
+COPY pkg/ pkg/
```
mgoodness, line 16: @wtam2018 I restored these lines because of the way Docker copies directories. `COPY api/ .`, for example, copies the contents of `api/` to the root of the workdir. We don't want that.

wtam2018: I thought that it was your intent to copy the entire folder to the root of WORKDIR.    It's fine to restore it.

wtam2018:
/LGTM
ty @mgoodness

OmerKahani:
@mgoodness I merged #96, do you need to test your branch with it, or is good to go?

mgoodness:
> @mgoodness, I merged #96. Do you need to test your branch with it, or is it good to go?

Good to go!

Issue #92, opened 2021-01-12
Intermittent test failure in 'TestGitGenerateParamsFromFiles/handles_error_during_getting_repo_file_contents'


jgwest:
The new test `TestGitGenerateParamsFromFiles/handles_error_during_getting_repo_file_contents` is intermittently failing locally and in GitHub actions.

See, for example, https://github.com/argoproj-labs/applicationset/actions/runs/479210059

### To reproduce:
```
./until_fail.sh  go test  -race -count=1 -timeout 30s -run ^TestGitGenerateParamsFromFiles$ github.com/argoproj-labs/applicationset/pkg/generators
```

Using a script file which runs the command over and over until it fails, `until_fail.sh`:
```
#!/bin/bash

while "$@"; do :; done
```

### Log

```
[jgw@localhost applicationset]$ ~/until-fail.sh /usr/local/go/bin/go test  -race -count=1 -timeout 30s -run ^TestGitGenerateParamsFromFiles$ github.com/argoproj-labs/applicationset/pkg/generators
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.329s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.328s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.324s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.316s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.318s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.303s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.300s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.307s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.348s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.341s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.306s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.307s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.279s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.307s
ok  	github.com/argoproj-labs/applicationset/pkg/generators	0.308s
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

[...]

PR #96, opened 2021-01-13
Intermittent test failure in 'TestGitGenerateParamsFromFiles/handles_error_during_getting_repo_file_contents'

jgwest:
The test failure was due to differing ordering of entries within a map between test runs; the fix is to use `assert.ElementsMatch(...)` as suggested by @mgoodness, plus ensuring a deterministic ordering in the order in which paths are processed by the generator (which end users of the generator should find to be a preferable behaviour).

Parent issue #92

### DIFF
--- a/pkg/generators/git.go
+++ b/pkg/generators/git.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"path"
+	"sort"
 	"time"
 
 	argoprojiov1alpha1 "github.com/argoproj-labs/applicationset/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


mgoodness (approved):

pkg/generators/git.go
```diff
@@ -76,20 +77,30 @@ func (g *GitGenerator) generateParamsForGitDirectories(appSetGenerator *argoproj
...
 	for _, requestedPath := range appSetGenerator.Git.Files {
 		paths, err := g.repos.GetPaths(context.TODO(), appSetGenerator.Git.RepoURL, appSetGenerator.Git.Revision, requestedPath.Path)
 		if err != nil {
 			return nil, err
 		}
 		for _, path := range paths {
-			allPaths[path] = true
+			allPathsMap[path] = true
 		}
 	}
```
OmerKahani, line 91: We can replace this whole block with https://godoc.org/github.com/mpvl/unique#Strings

we can collect all the paths to a list, and then call unique and sort on it

jgwest: My preference is to leave it as is rather than to introduce a new dependency. But if we find we do a lot of this throughout the code, we should reconsider.

wtam2018: Tough call.  I'd probably leave it as is.

OmerKahani: let's leave it for now, and handle it after the milestone.

OmerKahani (approved):

Issue #94, opened 2021-01-13
Setup release build scripts/artifacts for ApplicationSet controller releases


jgwest:
In order for us to perform an initial release of the ApplicationSet controller, we'll need build artifacts (scripts/makefile tweaks/etc) that do, at least:
- Performs pre-release check of git repository (see checks in sibling projects, below), container registry credentials
- Runs unit/E2E tests of current branch, exiting on failure
- Builds and pushes container image
- Regenerates the necessary Kubernetes manifests to point that image
- Creates a git commit containing the above
- Tags the commit with the release name, and pushes the release
- Finally, creates a corresponding GitHub release based on that tag
	
A number of the Argo/Argo-labs sibling projects have tackled this and have existing scripts we can likely borrow from, see for instance what [argocd-image-update](https://github.com/argoproj-labs/argocd-image-updater/blob/master/hack/release.sh) or [Argo CD](https://github.com/argoproj/argo-cd/blob/master/hack/trigger-release.sh) are doing as part of their release process.

Issue #70, opened 2020-12-09
The --namespace controller param and NAMESPACE environment variable should override to produce one canonical value


jgwest:
At present the `--namespace` controller parameter in `main.go`, and the `NAMESPACE` environment variable in `main.go` control different values:
- The `--namespace` param is passed to git generator
- The `NAMESPACE` is passed to `cache.MultiNamespacedCacheBuilder`

So:
- If both param and env are specified, one of them should override the other.
- That one value should then be used for both the git generator and the cache

ALSO: Need to update the Readme.md with this PR.

PR #105, opened 2021-01-21
Setup release build scripts/artifacts for ApplicationSet controller releases

jgwest:
The ApplicationSet controller release process and scripts in this PR are based on the process/scripts used by the [Argo CD Image Updater](https://github.com/argoproj-labs/argocd-image-updater), a sibling project in the Argo Labs organization. I've tweaked them slightly (adding support for quay.io, creating a release .md, allowing dry-run test using a personal repo, etc), but they are very similar otherwise.

I'm using 'quay.io' as the default release container registry, as that is what Argo CD itself has (recently) switched to.

As with Argo CD, and ArgoCD-Image-Updater, I am following the best practice of having a generated `manifests/install.yaml`, such that users can do a quick-start install using `kubectl apply -f manifests/install.yaml` for a release (this will be documented).
- The `manifest/install.yaml` in the PR is thus the generated version of this artifact for the current version

Closes https://github.com/argoproj-labs/applicationset/issues/94


hack/generate-manifests.sh
```diff
@@ -0,0 +1,37 @@
...
+  if [[ $branch = release-* ]]; then
+    pwd
+    IMAGE_TAG=v$(cat $SRCROOT/VERSION)
+  fi
+fi
+# otherwise, use latest
+if [ "$IMAGE_TAG" = "" ]; then
+  IMAGE_TAG=latest
+fi
+
```
wtam2018, line 32: We should add header comment "# This is an auto-generated file. DO NOT EDIT" to generated manifest to warn people.   Also, we will need CI validation to fail jobs that commit mismatch manifests (i.e. differ from generated manifests)

jgwest: @wtam2018  Done and done.

README.md
```diff
@@ -100,7 +100,7 @@ make build
 
 5. Run the Application Set Controller from the CLI:
 ```
-NAMESPACE=argocd ./dist/argocd-applicationset --metrics-addr=":18081" --probe-addr=":18082"
+NAMESPACE=argocd ./dist/argocd-applicationset --metrics-addr=":18081" --probe-addr=":18082" --argocd-repo-server=localhost:8081 --debug  --namespace=argocd
```
wtam2018, line 103: By adding `--namespaces=argocd`, we can elminite `NAMESPACE=argocd`?

manifests/crds/argoproj.io_applicationsets.yaml
```diff
@@ -475,6 +475,15 @@ spec:
                           - path
                           type: object
                         type: array
+                      files:
```
wtam2018, line 478: Just curious ... does this change related to the goal of the PR?

README.md
jgwest: Yah, I'm going to clean this up as part of https://github.com/argoproj-labs/applicationset/issues/70 , but until that bug is fixed both are needed.

manifests/crds/argoproj.io_applicationsets.yaml
jgwest: Since we've now added the `check-manifests` check above, this caused this file to fail due to being out-of-sync with the output of `make manifests`.

This file is likely out of sync due to a previous PR, and here I needed to clean it up in order to make `check-manifests` happy. Fortunately with this check in place, it should prevent that in the future.

jgwest: Looks like this same mismatch was also fixed in https://github.com/argoproj-labs/applicationset/pull/91, which I just merged, so this block is now no longer needed.

wtam2018 (approved):

PR #109, opened 2021-01-28
The --namespace controller param and NAMESPACE environment variable should override to produce one canonical value (#70)

### LINKED ISSUE #70: The --namespace controller param and NAMESPACE environment variable should override to produce one canonical value

jgwest:
Fixes #70 

As described on the issue, fix is to use the namespace value specified in this order (top is highest):
- CLI --namespace param
- NAMESPACE env var
- finally, default to 'argocd'

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -88,23 +88,29 @@ func main() {
...
+		namespace = os.Getenv("NAMESPACE")
+	}
+
+	// If neither the env var, nor the parameter are specified, use the Argo CD default
+	if len(namespace) == 0 {
+		namespace = "argocd"
 	}
-	setupLog.Info("using argocd namespace", "namespace", ns)
+
+	setupLog.Info("Using argocd namespace", "namespace", namespace)
```
OmerKahani, line 105: The log might be misleading if the namespace is not argocd

OmerKahani (approved):

PR #111, opened 2021-01-30
making examples more clear

### LINKED ISSUE #77: Move presently unimplemented examples out of examples/ root directory, update proposal links
The `examples/` directory of this repo contains a number of examples that are currently not implemented in the code (and are thus examples of how a developer *should implement in a future release*, not how to use for current users). This may be confusing for new users (and was confusing to me when initially onboarding to ApplicationSet controller), so we should move them to a subdir of examples/ and properly label them.

We can:
- Create a new `examples/proposed` directory (other suggestions for name/location are welcome!) that contains examples which are not yet implemented, leaving the root `examples/` directory for working examples. 
- Create a markdown file in the proposed directory saying the examples are not implemented in the code yet.
- The proposal doc contains direct links to these examples, and  we would should update the proposal doc to the new location. **Note**: We don't have edit access to the doc, only suggestion access, but my guess is one of the Argo CD principals should have edit access (perhaps Jesse?) and would be able to approve the changes.

Examples can "graduate" out of `proposed` as they are implemented.

OmerKahani:
Closes #77.

- Move filter to proposal.
- Delete the generic applicationset example; the use-case examples are more detailed
- Added values to the cluster example


jgwest (approved):
LGTM, thanks @OmerKahani!

Issue #35, opened 2020-09-02
Cluster generator cannot select local cluster


mgoodness:
The cluster generator currently requires a Kubernetes Secret to exist for each cluster. Argo CD doesn't use a Kubernetes Secret to identify the local cluster the way it does "external" clusters, so it can't be selected.

The easy workaround is to manually create a K8s Secret with the required `argocd.argoproj.io/secret-type=cluster` annotation and populate it with:

```
data:
  name: in-cluster
  server: https://kubernetes.default.svc
```

That Secret can then be annotated and labeled like any other, and the cluster-generator can use those values for templating.

However, that doesn't feel like a good solution in terms of usability.

mgoodness:
This is exactly what I'm doing pending a better solution:

```
apiVersion: v1
kind: Secret
metadata:
  labels:
    argocd.argoproj.io/secret-type: cluster
  name: cluster-local
  namespace: argocd
stringData:
  config: '{"bearerToken":"","tlsClientConfig":{"insecure":false,"caData":""}}'
  name: in-cluster
  server: https://kubernetes.default.svc
type: Opaque
```

jgwest:
Another workaround:
1) Within the Argo CD web UI, select **Settings**, then **Clusters**.
2) Select the your local cluster, usually named `in-cluster`.
3) Click the **Edit** button, and change the the **NAME** of the cluster to another value, for example `in-cluster-local`. Any other value here is fine. Leave all other fields unchanged.
4) Click **Save**.

The act of changing one of the default values for the local cluster causes the Argo CD Web UI will then create a new secret for this cluster. In the Argo CD namespace, you should now see a `Secret` resource named `cluster-(cluster suffix)` with label `"argocd.argoproj.io/secret-type": "cluster"`

OmerKahani:
Do we want to provide a solution for this, or are we good with the secret workaround?

One possible solution could be adding a boolean flag to the ClusterGenerator spec: `includeLocalCluster`.

jgwest:
I've got some work done on a branch, which I hope to get submitted sooner rather than later... when I do I'll post it here for your feedback.

jgwest:
Hey @mgoodness  @OmerKahani, I've posted a PR which now allows cluster generator to target the local cluster. See details on the PR as to how I'm doing it, and let me know what you think around the new behaviour.

morningspace:
[...]

PR #116, opened 2021-02-04
Cluster generator cannot select local cluster

jgwest:
This PR switches the behaviour of the cluster generator, such that it will now target both local AND non-local clusters (eg those with secrets defined).

If you wish to use the existing behaviour -- that is, targeting non-local clusters only with your Application -- then use a cluster selector with labels, for example:
```yaml
  generators:
  - clusters:
      selector:
        matchLabels:
          argocd.argoproj.io/secret-type: cluster
```

This selector will not match local clusters, since the local cluster does not have a secret (and thus does not have labels on that secret). Any cluster selector that specifies on a label will automatically exclude local clusters. 

If you wish to target both local and non-local cluster, using labels, you can create a secret for the local cluster as per the parent issue.

I'm using the existing Argo CD `util/db` package logic (pulled in directly from that package) to identify local/non-local clusters.

### Additional tweaks

Clean up cluster unit tests
- add test names
- remove base64 encoding on Data field (it should be decoded)
- general cleanup (add fields to test struct)

Clean up cluster E2E tests:
- move cluster e2e tests into their file
- miscellaneous test fixture tweaks

Fixes https://github.com/argoproj-labs/applicationset/issues/35


pkg/generators/cluster.go
```diff
@@ -52,24 +63,59 @@ func (g *ClusterGenerator) GenerateParams(
...
+	// ListCluster from Argo CD's util/db package will include the local cluster in the list of clusters
+	settingsMgr := settings.NewSettingsManager(g.ctx, g.clientset, g.namespace)
+	newDB := db.NewDB(g.namespace, settingsMgr, g.clientset)
+	clustersFromArgoCD, err := newDB.ListClusters(g.ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	if err := g.Client.List(context.Background(), clusterSecretList, client.MatchingLabelsSelector{Selector: secretSelector}); err != nil {
+	if clustersFromArgoCD != nil {
```
OmerKahani, line 80: we can do `if clustersFromArgoCD == nil { return nil,nil }`

pkg/generators/cluster.go
```diff
@@ -52,24 +63,59 @@ func (g *ClusterGenerator) GenerateParams(
...
+		clusters = clustersFromArgoCD.Items
+	}
+
+	clusterSecrets, err := g.getSecretsByClusterName(appSetGenerator)
```
OmerKahani, line 84: we can omit this function if `ignoreLocalClusters == false` because this means that there isn't any selector, and we will use all the items from clustersFromArgoCD

pkg/generators/cluster.go
jgwest: Done.

pkg/generators/cluster.go
jgwest: I believe the `getSecretsByClusterName` is still needed here, as it is used to distinguish local from non-local clusters later in the function body.

OmerKahani (approved):

PR #119, opened 2021-02-09
 Kustomize Argo CD integration (and GitHub E2E test action) should use a fixed Argo CD version

### LINKED ISSUE #118: Kustomize Argo CD integration (and GitHub E2E test action) should use a fixed Argo CD version
Using 'stable' as the Argo CD target in the [kustomize file](https://github.com/argoproj-labs/applicationset/blob/db51ff1d1be2f8f226826f0fb17c162272d90fc6/manifests/namespace-install-with-argo-cd/kustomization.yaml#L7) doesn't work well with our GitHub action that does manifest generation comparison (and fails on mismatch). The combination of these two means that any time Argo CD releases a new version (reasonably frequently), then all PRs/commits will fail the 'manifest diff' check until we update it.

The better solution is to target a fixed Argo CD version (consistent between both Kustomize and E2E tests) and bump it up at reasonable intervals.

jgwest:
This PR adds a check to manifest generation (via `hack/verify-argo-cd-versions.sh`), which verifies that all locations that reference Argo CD are consistent with a single Argo CD target version. 

This also:
- Updates to latest Argo CD v1.8.4
- Slightly cleans up `ci-build.yaml` and adds a new k3s-version for backwards compatibility testing (since I'm editing the file anyway)

Fixes #118

### DIFF
--- a/hack/verify-argo-cd-versions.sh
+++ b/hack/verify-argo-cd-versions.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# To adopt a new version of Argo CD:
+# 1) Update this value to the GitHub tag of the target 'argoproj/argo-cd' release (example: 'v1.8.4'). 
+# 2) Fix the errors that are reported below (by editing the version string in the file reported in the error)
+TARGET_ARGO_CD_VERSION=v1.8.4
+
+# Extract the Argo CD repository string from ci-build.yaml, which SHOULD contain the target Argo CD version
+VERSION_FROM_CI_BUILD=$( awk '/repository: argoproj\/argo-cd/,/Pin a/' .github/workflows/ci-build.yaml )
+
+if [[ $VERSION_FROM_CI_BUILD != *"$TARGET_ARGO_CD_VERSION"* ]]; then
+    echo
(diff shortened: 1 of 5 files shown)


hack/verify-argo-cd-versions.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# To adopt a new version of Argo CD:
+# 1) Update this value to the GitHub tag of the target 'argoproj/argo-cd' release (example: 'v1.8.4'). 
+# 2) Fix the errors that are reported below (by editing the version string in the file reported in the error)
+TARGET_ARGO_CD_VERSION=v1.8.4
+
+# Extract the Argo CD repository string from ci-build.yaml, which SHOULD contain the target Argo CD version
+VERSION_FROM_CI_BUILD=$( awk '/repository: argoproj\/argo-cd/,/Pin a/' .github/workflows/ci-build.yaml )
```
wtam2018, line 9: Does the awk script depends on the "# Pin a ..." command in ci-build.yaml?  If so, it would be safer to use some "marker" comment so that it does not get cleaned up or changed by other fellows.

jgwest: Done, thx @wtam2018

wtam2018 (approved):
LGTM
Thanks @jgwest

Issue #120, opened 2021-02-09
Create helm chart


maruina:
Hi all,
I see that Argo provides Helm charts to install ArgoCD or other tools (see https://github.com/argoproj/argo-helm/tree/master/charts). Are you planning to add an helm chart for the ApplicationSet as well? 

I'm happy to raise a PR. Should it be in the same ArgoCD chart as an option or a separate chart?

OmerKahani:
I think it can be part of the ArgoCD chart

maruina:
Ok, I will have a look and raise a PR there. Do you have a published docker image for the AppSet controller like https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml#L11?

jgwest:
I'd suggest a standalone ApplicationSet chart to begin with, as this would allow folks to install ApplicationSet alongside their existing Argo CD install.

We haven't put out an official 0.1.0 image out (still working our way through the [Milestone 1](https://github.com/argoproj-labs/applicationset/milestone/1) issues), but I've put out an image you can use for your testing:

I created a prerelease image `quay.io/argocdapplicationset/argocd-applicationset:v0.1.0-prerelease` here: 
https://quay.io/repository/argocdapplicationset/argocd-applicationset?tab=tags

The corresponding install artifacts are on my personal branch (https://github.com/jgwest/applicationset/tree/release-0.1.0):
```
Install ApplicationSet Controller:
kubectl apply -n argocd -f https://raw.githubusercontent.com/jgwest/applicationset/v0.1.0/manifests/install.yaml

Install ApplicationSet Controller + Argo CD v1.8.4:
kubectl create namespace argocd
kubectl apply -n argocd -f https://raw.githubusercontent.com/jgwest/applicationset/v0.1.0/manifests/install-with-argo-cd.yaml
```

As above, this is a temporary pre-release image, for testing purposes only. When we put out our first official release, this tag will be removed.

maruina:
Ok thanks, I'll raise a PR to add a separate applicationset helm chart and
then we can iterate from there. I will use the kustomize directory as a
base to create the chart. I should be able to start working on that in the
next few days.

maruina:
Started a draft PR: https://github.com/argoproj/argo-helm/pull/577

maruina:
@jgwest I'm tagging you here as well for visibility. I have the chart working but the tests are failing because of
```
[...]

PR #128, opened 2021-02-17
Use apiextensions.k8s.io/v1 CustomResourceDefinition, rather than deprecated v1beta1

### LINKED ISSUE #127: Use apiextensions.k8s.io/v1 CustomResourceDefinition, rather than deprecated v1beta1
The v1beta1 CRD definition is deprecated since Kubernetes v1.16, and is causing issues with the [Helm chart linting](https://github.com/argoproj/argo-helm/pull/577), also described [here](https://github.com/argoproj-labs/applicationset/issues/120), and previously caused issues during a kubebuilder scaffolding upgrade process.

Argo CD is going through the same process [here](https://github.com/argoproj/argo-cd/issues/5515).

jgwest:
The v1beta1 CRD definition is deprecated since Kubernetes v1.16, and is causing issues with the [Helm chart linting](https://github.com/argoproj/argo-helm/pull/577), also described [here](https://github.com/argoproj-labs/applicationset/issues/120).

This PR:
- Updates the makefile to generate v1 resources, rather than v1beta1
- Fixes issue with k8s behaviour in the `metadata` fields of a template, in generated CRD resource:
    - Since controller-gen [intentionally](https://github.com/kubernetes-sigs/controller-tools/blob/ea8d4ead3e34dac7bb792cedc6d3cd7decf63137/pkg/crd/known_types.go#L36) doesn't generate validation for `k8s.io/apimachinery/pkg/apis/meta/v1`, it is not included in the OpenAPI CRD schema. 
    - But, if using `apiextensions.k8s.io/v1` then k8s APIServer doesn't recognize this field as a valid metadata field (outside of it's use at the root). When you try to create an ApplicationSet with `.spec.template.metadata` fields (for example, from kubectl), the APIserver ignores the values.
     - Solution is to split the supported fields into their own ApplicationSet-managed type, rather than relying on ObjectMeta type.

Fixes #127


wtam2018 (approved):
LGTM
@jgwest thanks.

PR #132, opened 2021-02-19
Git Directory Generator only matches directories that contain valid Helm/ksonnet/Kustomize artifacts

### LINKED ISSUE #121: Git Directory Generator only matches directories that contain valid Helm/ksonnet/Kustomize artifacts
The Git Directory generator is currently calling Argo CD repo-server's `ListApps` function, in order to discover the list of directories that are available in the repository.

However, that function only returns directories that contain Argo CD applications (and it excludes directory-only applications), specifically those that [match the set of rules here](https://github.com/argoproj/argo-cd/blob/1d1fbf0fb17d5c47ff269b3232d090dc507e0b8d/util/app/discovery/discovery.go#L14), and [described in the docs here](https://argoproj.github.io/argo-cd/user-guide/tool_detection/).

While this works well for Argo CD, it doesn't fully fit our needs... as per the design doc, Git Directory should consider all directories, not just those that contain kssonet, Helm, or Kustomize artifacts.

So a couple better solutions are:
1) Either: Use `repo_service.go`'s `GetPaths` instead of `GetApps`, in order to gather the list of directories
2) Or: add a new function to Argo CD repo service to return a full list of paths.

IMHO I prefer option 1 (reserving the right to implement option 2 in the future, if we find that too much CPU/network resources are consumed by option 1).

jgwest:
See parent issue for details.

Fixes #121


pkg/services/repo_service.go
```diff
@@ -28,8 +29,16 @@ type argoCDService struct {
 }
 
 type Repos interface {
+	// GetApps return a list of valid Argo CD Application sources within the repo, as per the rules described here: https://argoproj.github.io/argo-cd/user-guide/tool_detection/
 	GetApps(ctx context.Context, repoURL string, revision string) ([]string, error)
```
OmerKahani, line 33: Is there any other use case of GetApps? If not, we can delete it.

jgwest: Agreed @OmerKahani , and done.

pkg/services/repo_service.go
```diff
@@ -26,8 +25,14 @@ type argoCDService struct {
 }
 
 type Repos interface {
-	GetApps(ctx context.Context, repoURL string, revision string) ([]string, error)
+
+	// GetPaths returns a list of files (not directories) within the target repo
 	GetPaths(ctx context.Context, repoURL string, revision string, pattern string) ([]string, error)
```
wtam2018, line 30: I wonder if this function should be renamed to `GetFilePaths`.

pkg/services/repo_service.go
```diff
@@ -87,12 +86,41 @@ func (a *argoCDService) GetPaths(ctx context.Context, repoURL string, revision s
...
 	}
 
-	paths, err := gitRepoClient.LsFiles(pattern)
-	if err != nil {
-		return nil, errors.Wrap(err, "Error during listing files of local repo")
+	filteredPaths := []string{}
+
+	repoRoot := gitRepoClient.Root()
+
+	if err := filepath.Walk(repoRoot, func(path string, info os.FileInfo, fnErr error) error {
```
wtam2018, line 93: Have you considered calling `LsFiles` and then use [filepath.Base()|https://golang.org/pkg/path/filepath/#Base] to only keep directores?   We won't have to explictly skip "." and .git.

jgwest: @wtam2018  - Good thinking, and actually I tried this approach first: the problem is that `LsFiles` doesn't list directories, only files, and there doesn't seem to be a switch to enable it show directories (see `man git-ls-files`). `LsFiles` is just a minimal wrapper around the `git ls-files --full-name -- (path to repo)` command, which you can try yourself on a local repo to see that it only lists files.

wtam2018 (approved):
LGTM
@jgwest  Thanks!

OmerKahani (approved):

Issue #71, opened 2020-12-09
ApplicationSet should report error conditions/status based on the result of generator/template resuls


jgwest:
When the user-provided generator/template produce invalid Argo CD `Application`s, we should note that error in the ApplicationSet CRD status field. See Argo CD's `ApplicationStatus` [conditions field](https://github.com/argoproj/argo-cd/blob/1ef14dca839423fe3cd9fade82bd1b3ed80eb5bd/pkg/apis/application/v1alpha1/types.go#L491).

Likewise I'm sure there are other status fields/conditions that would be useful to expose as part of an `ApplicationSet` `status` subtype.

Once implemented, one such function that would benefit from this is:
https://github.com/argoproj-labs/applicationset/issues/68

OmerKahani:
Hi, what do you think about this structure (based on argocd ApplicationStatus)
```
status:
  applications:
    - name:
      lastSync:
      condition:
        - type:
          message:
          LastTransitionTime
```

jgwest:
The way that Argo CD does conditions actually seems counter to current best practices around conditions: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

The API conventions page suggests the following condition fields:
```yaml
conditions:
  - type:  # Example: ErrorOccurred / ParametersGenerated / TemplateRendered
    status: # True/False/Unknown
    reason: # Single word camelcase representing the reason for the status eg MissingParameter or ErrorOccurred
    message: # Human readable message or error
    lastTransitionTime: # When this value last changed.
```

In addition to having conditions for each application, as you suggested, we can also have a conditions field at root the `status:` level too. This lets us communicate errors that aren't specific to particular applications. See the examples below.

This simulates an ApplicationSet with a single Application that failed on rendering the template:
```yaml
status:
  conditions:
    - type: ErrorOccurred
      status: True
      reason: MissingParameter
      message: "error occurred: Template referenced a parameter that doesn't exist" 
      lastTransitionTime: (...)
    - type: ResourcesUpToDate
      status: False
      reason: ErrorOccurred
      message: "A previous error occurred"
      lastTransitionTime:  (...)
  
  
  applications:
    - name: my-app
      condition:
        # Parameter generation was successful...
        - type: ParametersGenerated
          status: True
          reason: ""
          message: ""
          lastTransitionTime: (...)

        # But template rendering failed.
        - type: TemplateRendered
          status: False
          reason: MissingParameter
[...]

PR #136, opened 2021-02-25
Prevent ApplicationSet controller from creating invalid Applications, causing 'unable to delete application resource' in Argo CD

### LINKED ISSUE #112: Prevent ApplicationSet controller from creating invalid Applications, causing 'unable to delete application resource' in Argo CD
While writing a new E2E test, I happened to create an `ApplicationSet` resource that produced invalid Argo CD applictions:
- I created the ApplicationSet, and the ApplicationSet controller correctly created the Application.
- However, the created Application was invalid.
- Argo CD correctly complained that `application destination can't have both name and server defined: in-cluster`.

It was correct for Argo CD to complain about this, because *I did*, in fact, have both a name and a server defined in my ApplicationSet. The ApplicationSet I had defined was producing an invalid Application via the templating process.

The ApplicationSet controller should *never* create invalid Argo CD `Application`s (thus leaving it up to Argo CD to deal with that), it should instead detect if an `ApplicationSet` will produce an invalid `Application`, and rather than creating the `Application` it should instead log an error to the  console (and, in the future, report a message as a condition within the status field).

This invalid `Application` also prevented Argo CD from deleting the Application, with this message printed on delete (eg via the finalizer):
`
Unable to delete application resources: application destination can't have both name and server defined: in-cluster
`

This was the invalid Application that was produced:
```yaml
kind: Application
metadata:
  finalizers:
  - resources-finalizer.argocd.argoproj.io
  name: in-cluster-guestbook
  namespace: argocd-e2e
spec:
  destination:
    name: in-cluster
    namespace: guestbook2
  project: default
  source:
    path: guestbook
    repoURL: https://github.com/argoproj/argocd-example-apps.git
    targetRevision: HEAD
```

jgwest:
Fixes #112 

This PR:
- Use the Argo CD  "github.com/argoproj/argo-cd/util/argo" package's `ValidateDestination` and `ValidatePermissions` to verify generated application validity
- Adds unit tests to verify some of the desired validations

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -28,6 +28,8 @@ import (
 	"github.com/argoproj-labs/applicationset/pkg/generators"
 	"github.com/argoproj-labs/applicationset/pkg/utils"
 	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
+	"github.com/argoproj/argo-cd/util/argo"
+	"github.com/argoproj/argo-cd/util/db"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -43,6 +45,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	argoprojiov1alpha1 "github.com/argoproj-labs/applicationset/api/v1alpha1"
+	appclientset "github.com/argoproj/argo-cd/pkg/client/clientset/versioned"
+	argoutil "github.com/argoproj/argo-cd/util/argo"
+
+	apierr "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/imdario/mergo"
 )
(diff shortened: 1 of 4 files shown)


pkg/controllers/applicationset_controller.go
```diff
@@ -93,7 +102,7 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		// will fail. So just log it and return that the resource was
 		// successfully reconciled (which is true... it was reconciled to an
 		// error condition).
-		log.Errorf("ApplicationSet %s contains applications with duplicate name: %s", applicationSetInfo.Name, name)
+		log.Errorf("%s", validateError.Error())
 		return ctrl.Result{}, nil
```
OmerKahani, line 106: Do we want to fail if there is one invalid application? or we want best effort approach and apply the valid applications?

jgwest: Great question, and I thought about the same thing when creating the PR... I think either behaviour would be 'fine', but here was my thinking for the option I chose.

We have 2 options for how to handle this:
- **Option A:** apply all the valid applications, skip the invalid applications
- **Option B**: if at least one invalid application exists then don't apply any applications


I went with Option B, with this scenario in mind:

A user might be using a large ApplicationSet, with many different Argo CD Applications generated.

1) If one of those Applications is invalid, but 99 of them are valid, they may not notice the 1 invalid one. And at the moment (until more work [goes into this issue](https://github.com/argoproj-labs/applicationset/issues/71)), we don't have great error reporting to indicate problems.
    - So if we fail fast, and don't make any application changes, it is more likely the user will notice the lack of update and look for the issue. 

2) Likewise, individual Applications might have (implicit) dependencies on one another, and the user might not want all of the Application updates to be applied if one of the Applications is invalid. With Option B, the user must ensure that all generated templates render to valid Applications, before all the changes are enacted.


So that was my thinking, that our general philosophy should be to fail fast in cases where we know errors are present, thus limiting the blast radius of any potential failures. 

This should prevent potentially unwanted changes from being applied to the Argo CD managed cluster, with the slight inconvenience that the user needs to then debug the failing application in order to get the rest applied.

OmerKahani: that makes a lot of sense, thanks for the long answer!

OmerKahani (approved):

PR #148, opened 2021-03-05
add more cluster e2e tests

### LINKED ISSUE #80: Implement additional Cluster generator E2E tests
The proposed E2E test framework allows us to test ApplicationSet features against a live Kubernetes cluster and a live Argo CD instance. This issue tracks adding additional tests that specifically target the cluster generator.

PR #66 already includes a basic test of this generator, but this basic test was not an attempt at an exhaustive test of the functionality.

Should include, at least:
- Adding a new cluster (secret) should trigger ApplicationSet controller to create a new Application for it

OmerKahani:
Add two more tests:
- Adding a secret, after adding the applicationset
- Deleting a secret

closes #80

### DIFF
--- a/test/e2e/applicationset/cluster_e2e_test.go
+++ b/test/e2e/applicationset/cluster_e2e_test.go
@@ -220,3 +220,167 @@ func TestClusterGeneratorWithLocalCluster(t *testing.T) {
 		})
 	}
 }
+
+func TestSimpleClusterGeneratorAddingCluster(t *testing.T) {
+
+	expectedAppTemplate := argov1alpha1.Application{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Application",
+			APIVersion: "argoproj.io/v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "{{name}}-guestbook",
+			Namespace: utils.ArgoCDNamespace,
+		},
+		Spec: argov1alpha1.ApplicationSpec{
+			Project: "default",
+			Source: argov1alpha1.ApplicationSource{
+				RepoURL:        "https://github.com/argoproj/argocd-example-apps.git",
+				TargetRevision: "HEAD",
+				Path:           "guestbook",
+			},
+			Destination: argov1alpha1.ApplicationDestination{
+				Name:      "{{name}}",
+				Namespace: "guestbook",
+			},
+		},
+	}
+
+	expectedAppCluster1 := *expectedAppTemplate.DeepCopy()
+	expectedAppCluster1.Spec.Destination.Name = "cluster1"
+	expectedAppCluster1.ObjectMeta.Name = "cluster1-guestbook"
+
+	expectedAppCluster2 := *expectedAppTemplate.DeepCopy()
+	expectedAppCluster2.Spec.Destination.Name = "cluster2"
+	expectedAppCluster2.ObjectMeta.Name = "cluster2-guestbook"
+
+	Given(t).
+		// Create a ClusterGenerator-based ApplicationSet
+		When().
+		CreateClusterSecret("my-secret", "cluster1", "https://kubernetes.default.svc").
+		Create(v1alpha1.ApplicationSet{ObjectMeta: metav1.ObjectMeta{
+			Name: "simple-cluster-generator",
+		},
+			Spec: v1alpha1.ApplicationSetSpec{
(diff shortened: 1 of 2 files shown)


jgwest:
Hey @OmerKahani, I think what you are hitting is an Argo CD issue...

I have seen that if you:
1) Create a cluster named 'cluster1'
2) Create an application against that cluster
```yaml
kind: Application
metadata:
  name: cluster1-guestbook
  namespace: argocd-e2e
spec:
  destination:
    name: cluster1
    namespace: guestbook
  project: default
  source:
    path: guestbook
    repoURL: https://github.com/argoproj/argocd-example-apps.git
    targetRevision: HEAD
```
3) Delete that cluster secret

Then Argo CD seems to get stuck in a state where it keeps trying to process that Application over and over again (eg bounded only by network I/O bandwidth) in a reconciliation loop. Which means it never gets around to deleting it (and the test fails on missing deletion)

IMHO it seems like the correct behaviour is for Argo CD to  allow invalid Applications (Applications that point to clusters w/o a corresponding secret) to be deleted. The deletion finalizer and/or reconciler shouldn't prevent this operation.

OmerKahani:
> Hey @OmerKahani, I think what you are hitting is an Argo CD issue...
> 
> I have seen that if you:
> 
>     1. Create a cluster named 'cluster1'
> 
>     2. Create an application against that cluster
> 
> 
> ```yaml
> apiVersion: argoproj.io/v1alpha1
> kind: Application
> metadata:
>   name: cluster1-guestbook
>   namespace: argocd-e2e
> spec:
>   destination:
>     name: cluster1
>     namespace: guestbook
>   project: default
>   source:
>     path: guestbook
>     repoURL: https://github.com/argoproj/argocd-example-apps.git
>     targetRevision: HEAD
> ```
> 
>     1. Delete that cluster secret
> 
> 
> Then Argo CD seems to get stuck in a state where it keeps trying to process that Application over and over again (eg bounded only by network I/O bandwidth) in a reconciliation loop. Which means it never gets around to deleting it (and the test fails on missing deletion)
> 
> IMHO it seems like the correct behaviour is for Argo CD to allow invalid Applications (Applications that point to clusters w/o a corresponding secret) to be deleted. The deletion finalizer and/or reconciler shouldn't prevent this operation.

You just saved me hours of debugging :)

jgwest:
I investigated the Argo CD behaviour a bit more, and opened an Argo CD bug with steps to reproduce: https://github.com/argoproj/argo-cd/issues/5817

OmerKahani:
@jgwest maybe we should add a finalizer on the secret. Then the applicationset would delete the application prior to the deletion of the app. 

If the applicationset created the apps because of the secret, and the secret is been deleted I think that we should also delete the apps - what do you think?

jgwest:
Sorry for the delay review delay, @OmerKahani, in the process of reviewing and testing your changes, I discovered 


These tests have helped with:
- Helped test the workaround for https://github.com/argoproj-labs/applicationset/issues/170 and https://github.com/argoproj-labs/applicationset/pull/171
- Help find a race condition bug in the area of cluster creation/delete: https://github.com/argoproj-labs/applicationset/issues/176
- Help test a new PR to fix the above test (which is still only on my local, but which I will open on GH once 0.1.0 is out)

The code looks good to go, but I would like to hold off on merging it until https://github.com/argoproj-labs/applicationset/issues/176 is fixed (merging it noew would destabilize the E2E tests until the race condition is fixed.)

OmerKahani:
> Sorry for the review delay, @OmerKahani, in the process of reviewing and testing your changes I discovered and investigated race conditions in the cluster creation/deletion logic of ApplicationSet/ArgoCD, and your tests were super helpful in helping test them.
> 
> These tests have helped with:
> 
>     * Helped test the workaround for #170 and #171
> 
>     * Help find a race condition bug in the area of cluster creation/delete: #176
> 
>     * Help test a new PR to fix the above test (which is still only on my local, but which I will open on GH once 0.1.0 is out)
> 
> 
> So thanks for writing these! The code looks good to go, but I would like to hold off on merging it until #176 is fixed (merging it now would destabilize the E2E tests until the race condition is fixed.)

Thank you!

These are probably the best tests I had ever written :laughing:

jgwest (approved):
As mentioned above, looks great and very helpful for finding bugs! Thanks for taking this on, and thanks for your patience in the time it took to get it merged :smile: .

PR #156, opened 2021-03-15
Annotations set in application-set not updating apps

### LINKED ISSUE #154: annotations set in application-set not updating apps
```
  template:
    metadata:
      annotations:
        argocd.argoproj.io/manifest-generate-paths: consul/consul-agent
```
With application annotations it will create the annotation correctly but if you try to modify it the changes dont sync unless you delete the application and regenerate. Pretty easy to test, create a annotation like above and then try changing it.

jgwest:
Fixes #154

The fix for this bug is the change to `createOrUpdateInCluster` in `pkg/controllers/applicationset_controller.go`, everything else in this PR is test updates to ensure the correctness of the fix (for both unit and E2E tests).

The bug occurred because:
- The mutateFunction of `createOrUpdateInCluster` was not copying metadata fields.
- The call to `c.Get(...)` in `CreateOrUpdate(...)` in `createOrUpdate.go` was not replacing the existing fields of the Application parameter that was passed in, causing the objects to appear equal (when they were not). Fix is to pass in a preinitialized `argov1alpha1.Application{}` value into `utils.CreateOrUpdate(...)`. (It's unclear whether this is only an issue with the fakeclient, or a general problem, but in any case this is the fix.)

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -407,21 +407,36 @@ func (r *ApplicationSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // createOrUpdateInCluster will create / update application resources in the cluster.
-// For new application it will call create
-// For application that need to update it will call update
-// The function also adds owner reference to all applications, and uses it for delete them.
+// - For new applications, it will call create
+// - For existing application, it will call update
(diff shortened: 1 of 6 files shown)


wtam2018 (approved):
LGTM
thanks @jgwest !

PR #161, opened 2021-03-17
Git helpers missing from image

### LINKED ISSUE #160: git helpers missing from image
When using credentials for git repositories the repo service is looking for binaries such as: https://github.com/argoproj/argo-cd/blob/master/hack/git-ask-pass.sh

jgwest:
As reported on the parent issue, the `.sh` files that Argo CD uses for its git CLI calls are missing from the ApplicationSet image.  This causes a blocking `Username for 'https://github.com':` message from the controller.

Fix is to add those `.sh` files in the appropriate spot within the image.

(I also had to tweak the `.github/workflows/ci-build.yaml` because kustomize version in the GitHub runner was upgraded recently, and we should have been using a fixed version to begin with.)

Fixes #160

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -44,7 +44,10 @@ jobs:
         uses: actions/checkout@v2
       - name: Regenerate the manifests, git diff should report no changes.
         run: |
-          make manifests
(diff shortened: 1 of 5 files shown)


wtam2018 (approved):
LGTM
thanks @jgwest !

PR #171, opened 2021-03-25
Workaround Argo CD cluster deletion reconciliation bug

### LINKED ISSUE #170: Workaround Argo CD cluster deletion reconciliation bug
Argo CD bug https://github.com/argoproj/argo-cd/issues/5817 causes significant issues for the Cluster generator of the ApplicationSet controller: when a cluster is deleted from Argo CD, and that cluster had Applications associated with it, Argo CD is unable to delete the Application and the application controller enters a reconciliation loop.

This is primarily an Argo CD bug, but is easily reproducible with the ApplicationSet controller (due to us supporting automated deletion of Applications on cluster deletion), hence the need to workaround it until it is fixed in v1.8 and v2.x Argo CD versions.

jgwest:
This PR removes the Application finalizer if the cluster destination for that Application is detected to be is invalid (for example, the cluster was removed from Argo CD, or the cluster secret was deleted). This ensures that the Application can be deleted successfully without locking up the Argo CD instance (See the parent issue for details on Argo CD's bad behaviour).

Since the cluster destination does not exist, there are no application resources to clean up, so we are safe to remove the finalizer (but until the upstream bug is fixed in supported Argo CD versions, we need to workaround it here).

Fixes #170


OmerKahani:
Do we need unit tests, or can we really on the E2E tests?

pkg/controllers/applicationset_controller.go
```diff
@@ -520,6 +520,28 @@ func (r *ApplicationSetReconciler) deleteInCluster(ctx context.Context, applicat
...
+
+			// Only check if the finalizers need to be removed IF there are finalizers to remove
+			if len(app.Finalizers) > 0 {
+				// If the destination is invalid (for example the cluster is no longer defined), then remove
+				// the application finalizers to avoid triggering Argo CD bug #5817
+				if err := argoutil.ValidateDestination(ctx, &app.Spec.Destination, r.ArgoDB); err != nil {
+					app.Finalizers = []string{}
+
+					r.Recorder.Eventf(&applicationSet, corev1.EventTypeNormal, "Updated", "Updated Application %q finalizer before deletion", app.Name)
+					appLog.Log(log.InfoLevel, "Updating application finalizer before deletion")
```
OmerKahani, line 532:
```suggestion
					appLog.Log(log.InfoLevel, "Updating application finalizer before deletion because application has an invalid destination")
```

pkg/controllers/applicationset_controller.go
```diff
@@ -520,6 +520,28 @@ func (r *ApplicationSetReconciler) deleteInCluster(ctx context.Context, applicat
...
 		if !exists {
+
+			// Only check if the finalizers need to be removed IF there are finalizers to remove
+			if len(app.Finalizers) > 0 {
+				// If the destination is invalid (for example the cluster is no longer defined), then remove
+				// the application finalizers to avoid triggering Argo CD bug #5817
+				if err := argoutil.ValidateDestination(ctx, &app.Spec.Destination, r.ArgoDB); err != nil {
+					app.Finalizers = []string{}
+
+					r.Recorder.Eventf(&applicationSet, corev1.EventTypeNormal, "Updated", "Updated Application %q finalizer before deletion", app.Name)
```
OmerKahani, line 531:
```suggestion
					r.Recorder.Eventf(&applicationSet, corev1.EventTypeNormal, "Updated", "Updated Application %q finalizer before deletion, because application has an invalid destination", app.Name)
```

pkg/controllers/applicationset_controller.go
```diff
@@ -520,6 +520,28 @@ func (r *ApplicationSetReconciler) deleteInCluster(ctx context.Context, applicat
 		_, exists := m[app.Name]
 
 		if !exists {
+
+			// Only check if the finalizers need to be removed IF there are finalizers to remove
+			if len(app.Finalizers) > 0 {
```
OmerKahani, line 525: we should verify that the finalizer here is "resources-finalizer.argocd.argoproj.io" and only delete that one

OmerKahani (changes requested):

pkg/controllers/applicationset_controller.go
jgwest: Done.

pkg/controllers/applicationset_controller.go
jgwest: Done.

pkg/controllers/applicationset_controller.go
jgwest: Done.

jgwest:
Thanks @OmerKahani , I've updated the PR with your requested changes, and add unit tests for the new method. If you can take a look ASAP that would be great, as I'd like to update the RC driver with this change so it can be tested. Thanks :smile: !

OmerKahani (approved):

OmerKahani:
> Thanks @OmerKahani , I've updated the PR with your requested changes, and add unit tests for the new method. If you can take a look ASAP that would be great, as I'd like to update the RC driver with this change so it can be tested. Thanks 😄 !

Looks great now!

wtam2018:
LGTM
We are going to remove the workaround when the parent bug is resolved, right?   What will remind us to do that?

jgwest:
Thanks @wtam2018 :+1: , opened https://github.com/argoproj-labs/applicationset/issues/175 to remind us, and added it to next milestone.

PR #187, opened 2021-04-06
[WIP] Add Merge Patch option for ApplicationSet sync

### LINKED ISSUE #186: ApplicationSet Controller overrides any Application CR changes
Related to #180.

Right now if you attempt to add annotations, labels, or do things like per-app parameter overrides, those changes are automatically and quickly overridden. It'd be nice if (either via config or by default) the controller merged the generated CR instead of overwriting it.

RichiCoder1:
Adds "MergePatch" sync option for ApplicationSets when upserting generated sets.

Closes #186


RichiCoder1:
Closing this as essentially what needs to happen is the generated template is stored (in an annotation) on the generated app, and updated app is compared to the previously generated app and a patch is generated, and then the patch is applied to the live resource. Basically a three-way merge.

allupaku:
Is this feature available in any releases, I cannot find it.

Issue #176, opened 2021-03-29
Race condition on cluster creation/deletion causes unexpected behaviour


jgwest:
On cluster creation/deletion, I have seen cases where:
- If a new Argo CD cluster is created, and the ApplicationSet is reconciled before the 'new cluster event' is received by the Argo CD DB cluster lister, then a new Application is failed to be created for the cluster.
- If a new Argo CD cluster is deleted, and the ApplicationSet is reconciled before the 'cluster deleted event' is received by the Argo CD DB cluster lister, then the Application fails to be deleted.

This can be reproduced by running the E2E tests that @OmerKahani implemented as part of https://github.com/argoproj-labs/applicationset/pull/148 (eg the tests that creates a new cluster secret).

I've got a PR for this, but I'm targeting this for the next milestone because I think the fix would be too invasive at this late stage for the 0.1.0 release, because it is partially mitigated by https://github.com/argoproj-labs/applicationset/issues/170, and because it should occur infrequently in real world settings.

I was reminded of this issue that I opened, which might be related: https://github.com/argoproj/argo-cd/issues/4755
- I could see this as being the Argo CD equivalent of this ApplicationSet issue

## Log

When running `make test-e2e` with Omer's E2E PR, it's reproducible about 20% of the time

```
=== RUN   TestSimpleClusterGeneratorAddingCluster
time="2021-03-25T12:32:25-04:00" level=info msg="mkdir -p /tmp/applicationset-e2e" dir= execID=ey9ig
time="2021-03-25T12:32:25-04:00" level=debug duration=2.24523ms execID=ey9ig
time="2021-03-25T12:32:25-04:00" level=info msg="clean state" duration=46.269361ms id= name=TestSimpleClusterGeneratorAddingCluster password=password username=admin
time="2021-03-25T12:32:26-04:00" level=info msg="action: creating cluster Secret 'my-secret'"
time="2021-03-25T12:32:26-04:00" level=info msg="expectation succeeded: no error and output contained ''"
time="2021-03-25T12:32:26-04:00" level=info msg="action: creating ApplicationSet 'simple-cluster-generator'"
time="2021-03-25T12:32:26-04:00" level=info msg="expectation succeeded: no error and output contained ''"
time="2021-03-25T12:32:26-04:00" level=info msg="expectation pending: missing app 'cluster1-guestbook'"
time="2021-03-25T12:32:29-04:00" level=info msg="expectation succeeded: all apps successfully found"
[...]

PR #198, opened 2021-04-12
Race condition on cluster creation/deletion causes unexpected behaviour 

jgwest:
See details on parent issue.

The vast majority of this PR is test changes, the new files in `pkg/utils` are the only substantial changes to product code.

This PR:
- Forks a few of the `util/db/cluster.go` functions from Argo CD, and replaces their use of the race-condition-prone `ApplicationLister`, with a simple kubeclient query.
- Forks the test cases that test those functions, as well, to make sure nothing has regressed.
- As I state in the PR, ideally we would upstream these changes to Argo CD in some form, but in the mean time (and until Argo CD has adopted the fix in all the versions we support), this fixes the issue here.

Fixes https://github.com/argoproj-labs/applicationset/issues/176

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	k8s.io/api v0.19.2
 	k8s.io/apimachinery v0.19.2
 	k8s.io/client-go v11.0.1-0.20190816222228-6d55c1b1f1ca+incompatible
+	k8s.io/utils v0.0.0-20200912215256-4140de9c8800
 	sigs.k8s.io/controller-runtime v0.7.0
 )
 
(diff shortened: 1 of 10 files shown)


OmerKahani:
What would be the steps once ArgoCD adopts the fix? Do we revert this PR?

OmerKahani (approved):

jgwest:
@OmerKahani Good question, what we need from Argo CD is a version of this API (in the Argo CD code) that doesn't use the Secret lister (which can have delayed results that cause the race condition), but instead just query the cluster directly.

Once this is in place in Argo CD, we would pull in that new version of Argo CD as a dependency, and switch to using that API. At that point, we would:
- Delete `pkg/utils/cluster*.go` (no longer needed)
- Replace references to `"github.com/argoproj-labs/applicationset/pkg/utils"` with the new Argo CD package

PR #201, opened 2021-04-15
Ignore folders start with "."

LiorLieberman:
Make repo service ignore all folders starts with "."


jgwest (approved):
LGTM, short and sweet, thanks @LiorLieberman !

PR #204, opened 2021-04-19
Correct list generator syntax

gdubya:
The list generator calls them "elements" rather than "items"


jgwest (approved):
Thanks @gdubya , good catch, this is the old syntax.

PR #205, opened 2021-04-21
Add Matrix generator

### LINKED ISSUE #114: feature request - combine generators
Hi folks,

first, thanks for this great project!

Would it be possible to combine generators?

in this example combine directory discover with clusters secret

example:

```
kind: ApplicationSet
metadata:
  name: cluster-addons
spec:
  generators:
  - git:
      repoURL: https://github.com/infra-team/cluster-deployments.git
      directories:
      - path: add-ons/*
  - clusters:
     selector:
       matchLabels:
         argocd.argoproj.io/secret-type: cluster
       values:
         project: default 
  template:
    metadata:
      name: '{{name}}-{{path.basename}}'
    spec:
      project: '{{values.project}}'
      source:
        repoURL: https://github.com/infra-team/cluster-deployments.git
        targetRevision: HEAD
        path: '{{path}}'
      destination:
        server: {{server}}
        namespace: '{{path.basename}}'
```

When I try, run into
```
error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"server":interface {}(nil)}
```

OmerKahani:
fix #114


jgwest:
@OmerKahani 

Nice, I presume it should be `matrix` rather than `metrix`?

Re: the verify generated manifests GitHub action failure, you probably need to make sure to use Kustomize v3.9.4 (Lior hit the same issue here: https://argoproj.slack.com/archives/D01UD22L081/p1618251794009300)... I have an idea of an easy way to fix this issue, but in the mean time, as a workaround, Kustomize v3.9.4 should solve the problem.

jgwest:
Opened an issue for the 'verify generated manifests' issue described above: https://github.com/argoproj-labs/applicationset/issues/206

OmerKahani:
Still doing some refactoring, and adding more docs - but it would be great to get feedback

docs/Generators.md
```diff
@@ -370,3 +370,63 @@ Any `config.json` files found under the `cluster-config` directory will be param
...
+```
+3. The matrix generator ignores templates of the inner generators
+```yaml
+- matrix:
+    generators:
+        - list:
+            elements: []
+            template: # Ignored
+       
+```
```
jgwest, line 432: It couldn't hurt to give an example, something like:

### Example

Imagine we have two clusters:
- cluster A (@ https://1.2.3.4)
- cluster B (@ https://2.4.6.8)

And two git directories in our git repository:
- /examples/git-generator-directory/cluster-addons/argo-workflows  (argo workflows controller)
- /examples/git-generator-directory/cluster-addons/prometheus-operator (prometheus operator)

```
cluster generator generates these parameters:
  - generate parameters:
    - name: clusterA
      server: https://1.2.3.4
    - name: clusterB
      server: https://2.4.6.8


git directory generator:
  - generates parameters:
    - path: /examples/git-generator-directory/cluster-addons/argo-workflows
      path.basename: argo-workflows
      
    - path: /examples/git-generator-directory/cluster-addons/prometheus-operator
      path.basename: prometheus-operator
```

Which will become these sets of parameters:
```
    - name: clusterA
      server: https://1.2.3.4
      path: /examples/git-generator-directory/cluster-addons/argo-workflows
      path.basename: argo-workflows

    - name: clusterA
      server: https://1.2.3.4
      path: /examples/git-generator-directory/cluster-addons/prometheus-operator
      path.basename: prometheus-operator

    - name: clusterB
      server: https://2.4.6.8
      path: /examples/git-generator-directory/cluster-addons/argo-workflows
      path.basename: argo-workflows



    - name: clusterB
      server: https://2.4.6.8      
      path: /examples/git-generator-directory/cluster-addons/prometheus-operator
      path.basename: prometheus-operator
```
This allows you to target each cluster with the applications specified at the specific git repositories.

api/v1alpha1/applicationset_types.go
```diff
@@ -66,6 +66,16 @@ type ApplicationSetGenerator struct {
 	List     *ListGenerator    `json:"list,omitempty"`
 	Clusters *ClusterGenerator `json:"clusters,omitempty"`
 	Git      *GitGenerator     `json:"git,omitempty"`
+	Matrix   *MatrixGenerator  `json:"matrix,omitempty"`
+}
+
+// ApplicationSetBaseGenerator include list item info
+// CRD dosn't support recursive types so we need a different type for the matrix generator
```
jgwest, line 73: Wow, I didn't know this, TIL :smile:

docs/Generators.md
```diff
@@ -370,3 +370,63 @@ Any `config.json` files found under the `cluster-config` directory will be param
...
+    spec:
+      project: '{{metadata.labels.environment}}'
+      source:
+        repoURL: https://github.com/argoproj-labs/applicationset.git
+        targetRevision: HEAD
+        path: '{{path}}'
+      destination:
+        server: '{{server}}'
+        namespace: '{{path.basename}}'
+```
```
jgwest, line 411: Looks like the indentation on the example is slightly off, `generators:` (and everything under it) should be shifted right by 2 spaces, since `generators` is a field of `matrix`.
Also it looks like the `dependsOn: cluster-deployments` field shouldn't be there, as this was only part of the previous examples (correct me if I am wrong)

docs/Generators.md
```diff
@@ -370,3 +370,63 @@ Any `config.json` files found under the `cluster-config` directory will be param
...
+       git: ...
+```
+3. The matrix generator ignores templates of the inner generators
+```yaml
+- matrix:
+    generators:
+        - list:
+            elements: []
+            template: # Ignored
+
```
jgwest, line 431: Whitespace likewise looks slightly off in these examples, re:  to the left of `- list:` in 2 and 3.

docs/Generators.md
```diff
@@ -370,3 +370,63 @@ Any `config.json` files found under the `cluster-config` directory will be param
...
+        targetRevision: HEAD
+        path: '{{path}}'
+      destination:
+        server: '{{server}}'
+        namespace: '{{path.basename}}'
+```
+(*The full example can be found [here](https://github.com/argoproj-labs/applicationset/tree/master/examples/matrix).*)
+
+### Limitations
+1. The matrix generator support only two inner generators
```
jgwest, line 415: It looks like we only support two inner generators (eg 2 dimensional matrix):
- I presume we could theoretically support more in the future (eg n-dimensional matrix, unless there is some technical or strategic reason not to)?
- It looks like we would only need to update the generation algorithm to support a dynamic number of elements, and add tests (but we can look at that later, if/when someone asks for it)

In any case, think we should change the line to `The matrix generator support only two inner generators at this time.` or `The matrix generator currently supports only two inner generators.`

docs/Generators.md
```diff
@@ -370,3 +370,63 @@ Any `config.json` files found under the `cluster-config` directory will be param
...
+        repoURL: https://github.com/argoproj-labs/applicationset.git
+        targetRevision: HEAD
+        path: '{{path}}'
+      destination:
+        server: '{{server}}'
+        namespace: '{{path.basename}}'
+```
+(*The full example can be found [here](https://github.com/argoproj-labs/applicationset/tree/master/examples/matrix).*)
+
+### Limitations
```
jgwest, line 414: What do you think about `### Restrictions` rather than Limitations as a header name?

examples/matrix/cluster-and-git.yaml
```diff
@@ -0,0 +1,34 @@
...
+#
+#
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cluster-git
+spec:
+  generators:
+  - matrix:
+    generators:
```
jgwest, line 12: Indentation is off in these examples, it looks like `generators:`, and all its children, should be shifted to the right by 2.

examples/matrix/cluster-and-git.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - git:
+          name: cluster-deployments
+          repoURL: https://github.com/argoproj-labs/applicationset.git
+          directories:
+          - path: examples/proposal/matrix/cluster-addons/*
+      - clusters:
+          dependsOn: cluster-deployments
```
jgwest, line 19: `dependsOn` line can be removed.

examples/matrix/cluster-and-git.yaml
```diff
@@ -0,0 +1,34 @@
+# This example demonstrates the combining of the git generator with a clusgter generator
```
jgwest, line 1: A couple typos in lines 1 and 2, here.

examples/matrix/list-and-git.yaml
```diff
@@ -0,0 +1,40 @@
+# This example demonstrates the combining of the git generator with a list generator
+# The expected output would be an application per git directory and a list entry (applicztions_count = git directory * list entries)
+#
+#
```
jgwest, line 4: See comments on previous example.

pkg/generators/utils.go
```diff
@@ -0,0 +1,15 @@
...
+	for k, v := range a {
+		res[k] = v
+	}
+
+	for k, v := range b {
+		res[k] = v
+	}
+
+	return res
+}
```
jgwest, line 15: I think we should return an error if there are duplicate keys detected (eg both generators happen to generate the same parameter name), unless the values are also the same. I think this behaviour is preferable versus using the reverse order of generators in the list to determine which duplicate key to use (which would be the unpatched  behaviour).

pkg/generators/matrix.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+var _ Generator = (*MatrixGenerator)(nil)
+
+var MoreThanTwoGenerators = errors.New("found more than two generators, Matrix support only two")
+var LessThanTwoGenerators = errors.New("found less than two generators, Matrix support only two")
+var MoreThenOneInnerGenerators = errors.New("found more than one generator in matrix.Generators")
+
+type MatrixGenerator struct {
+	generators map[string]Generator
```
jgwest, line 18: I wasn't sure what this field was until I read through the code, what if we change the name of the  'generators' field to `supportedGenerators` with a comment like 'these are interfaces of all the other types of generators supported (cluster, list, git)'

pkg/generators/matrix.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if len(t) > 1 {
+		return nil, MoreThenOneInnerGenerators
+	}
+
+	return t[0].Params, nil
+}
+
+func (m *MatrixGenerator) GetRequeueAfter(appSetGenerator *argoprojiov1alpha1.ApplicationSetGenerator) time.Duration {
+	return NoRequeueAfter
```
jgwest, line 77: I think for Matrix generator's `GetRequeueAfter`, it should call `GetRequeueAfter` on all the inner generators, and returned the largest value received? Otherwise we might be suppressing a non-zero RequeueAfter value from an inner generator (unless this is handled elsewhere).

jgwest (changes requested):
Looks really great @OmerKahani  :smile:, I really like what you've done here... I found some minor tweaks I've noticed, let me know what you think!

api/v1alpha1/applicationset_types.go
OmerKahani: This was really painful to find out

jgwest: Yah, I can image :smile: :frowning:

docs/Generators.md
OmerKahani: you are correct, I wanted to start with two as the code is simpler and the PR would be easier (it's a big PR as is)

docs/Generators.md
OmerKahani: Add the example, what do you think?

docs/Generators.md
```diff
@@ -369,3 +369,113 @@ spec:
...
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cluster-git
+spec:
+  generators:
+    - matrix:
+        generators:
+            - git:
+                name: cluster-deployments
```
jgwest, line 400: No `name:` field on Git generator, think this can be removed? Plus need to add the `revision` field which is required.

examples/matrix/list-and-git.yaml
```diff
@@ -0,0 +1,40 @@
...
+  generators:
+  - matrix:
+    generators:
+      - git:
+          name: cluster-deployments
+          repoURL: https://github.com/argoproj-labs/applicationset.git
+          directories:
+          - path: examples/proposal/matrix/cluster-addons/*
+      - list:
+          dependsOn: cluster-deployments
```
jgwest, line 19: `dependsOn` can be removed.

examples/matrix/cluster-and-git.yaml
```diff
@@ -0,0 +1,33 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cluster-git
+spec:
+  generators:
+  - matrix:
+      generators:
+        - git:
+            name: cluster-deployments
```
jgwest, line 14: Need to add the `revision` field to `git:`, which is required.

jgwest (changes requested):
Looking good @OmerKahani , I found some example tweaks that are required, and it looks like you'll need to rebase, but otherwise good to go.

winmillwill:
Should we also change parameter keys from different generators so that they don't collide? For example, right now both scmProvider and list generators will try to set `url`. I have a use case for generating a matrix that is the product of arbitrary values from the list generator and all the repos in my org with viable kube manifests at a given path.

OmerKahani:
> Should we also change parameter keys from different generators so that they don't collide? For example, right now both scmProvider and list generators will try to set `url`. I have a use case for generating a matrix that is the product of arbitrary values from the list generator and all the repos in my org with viable kube manifests at a given path.

this will be a breaking change, so maybe would be best handled outside this PR (which is big as is)

stecky:
Looks like we are getting close?!?! 🎉

jgwest (approved):
Good to go :tada: :tada: :tada:! (FYI I pushed a commit to remove the merge conflict in the docs between scmProvider and matrix generator, hopefully nothing is missing from either :smile: )

jgwest:
Great suggestion @winmillwill (and FYI @OmerKahani too), I've opened up https://github.com/argoproj-labs/applicationset/issues/236 to continue the discussion and posted some thoughts. 

(@winmillwill I believe you will need to manually subscribe to that issue if you wish to get notifications of new comments)

PR #207, opened 2021-04-21
Use a fixed version of Kustomize in the Makefile to generate manifests

### LINKED ISSUE #206: Use a fixed version of Kustomize in the Makefile to generate manifests
Since a newer version of kustomize has been released since [we last set a fixed kustomize version](https://github.com/argoproj-labs/applicationset/pull/161), this is causes the 'verify manifests' job of GitHub actions to fail on new PRs.

We need to ensure that the Makefile is always run with the correct Kustomize version.

Kubebuilder/operator-sdk scaffolds a controller with a fixed kustomize version, which is automatically downloaded under `(project root)/bin`. We can use the same Makefile scaffolding here, to download the desired version of kustomize and use it with kustomize calls.

jgwest:
Fixes #206

See parent issue for details.

Makefiles generated by kubebuilder/operator-sdk include a section of that automatically downloads a specific `kustomize` version. Here I've borrowed that section and hooked it into our Makefile's logic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ MKDOCS_RUN_ARGS?=
 
 CURRENT_DIR=$(shell pwd)
 
+KUSTOMIZE = $(shell pwd)/bin/kustomize
+
 
 ifdef IMAGE_NAMESPACE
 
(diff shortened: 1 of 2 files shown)


OmerKahani (approved):
Thanks!

Makefile
```diff
@@ -47,15 +49,15 @@ image-push: image
 	docker push ${IMAGE_PREFIX}${IMAGE_NAME}:${IMAGE_TAG}
 
 .PHONY: deploy
-deploy: manifests
-	kustomize build manifests/namespace-install | kubectl apply -f -
+deploy: kustomize manifests
```
wtam2018, line 52: It looks like `manifests` depends on `kustomize` already so we don't need an explicit `kustomize` dependency here.

Issue #103, opened 2021-01-20
Method for overriding resource deletion default behavior


jbaskeen:
Currently, the AppSet spec enforces deletion of related Applications and their deployed resources by default. This is not consistent with the default behavior of Application in argo-cd. That is, without AppSets, deployed resources related to an Application are not deleted by default when an Application is deleted. This situation presents a rather confusing situation for users of Application _and_ ApplicationSet since their default behavior for resource deletion has diverged.

Proposal

- Allow for a method to set the default behavior of ApplicationSet to not delete resources when an ApplicationSet is deleted. A runtime flag would be a possible solution to set the desired default behavior.
- Allow for the controller's default behavior to be overridden or set within the ApplicationSet spec via an attribute setting.

Justification

In some cases, critical infrastructure resources or applications may be synced to a cluster using ApplicationSet which need to be "protected" from deletion in case of human error in ArgoCD configurations. In the event that an unintentional deletion, by whatever means, of AppSets, or Applications that are linked to AppSets, a Production environment state could be lost causing an outage of provided services. This exact scenario occurred on my team where an inappropriate Argo configuration caused deletion of our AppSets resulting in all environments to be wiped clean by Argo. This occurred because of the default behavior of ApplicationSet with respect to deletions and we did not expect that behavior (we expected the same behavior as Application) as a result of our lack of due diligence prior to Production usage. In this specific case, persistent storage which was provisioned via Argo was also lost. In the case of storage, as a sensitive asset type, it is reasonable to expect a way of protecting those resources from erroneous deletion because of their stateful nature.

jgwest:
@jbaskeen has provided an excellent description of why this is a valid use case, and I am personally convinced that we should allow users to customize this behaviour if they wish (eg as a supported behaviour). 

#### Re: changing the default behaviour

IMHO the current default for ApplicationSet is the correct default: ApplicationSet is meant to be an abstraction over Argo CD `Application` resources, with the user managing `ApplicationSet` resources directly, and the ApplicationSet controller handling changes to the corresponding `Application`s behind the scenes.

Thus, to preserve that abstraction as much as possible, the lifecycle of the ApplicationSet, of the Application, and of the Application resources, should be equal. Eg when the ApplicationSet is deleted, so is the Application, and so are the Application's resources. 

(Likewise, I would note that Argo CD doesn't have a consistent default here: for example, the web UI and the CLI both use cascade deletion by default)

#### Proposal

[...]

PR #223, opened 2021-05-04
feat: preserve resources on deletion

mmatur:
### Description

This PR is an implementation to preserve resource when an applicationset is deleted.

Fixes: #103


jgwest (approved):
LGTM, thanks @mmatur!

haslersn:
This is a critically important feature and should be released asap.

agaudreault-jive:
Is there a way to set `preserveResourcesOnDeletion: true` as the default behavior in the applicationSet controller ? If not, do you think that would be a good feature ?

PR #233, opened 2021-05-14
Use a fixed version of controller-gen in the Makefile to generate manifests

### LINKED ISSUE #232: Use a fixed version of controller-gen in the Makefile to generate manifests
The Makefile we're using to generate the ApplicationSet manifests (and the CI checks that check the resulting artifacts) shouldn't assume a specific version of `controller-gen` is installed on the user's path.

This is the same type of fix we previously made for `kustomize` on https://github.com/argoproj-labs/applicationset/issues/206. We will download the appropriate `controller-gen` version to `(applicationset project root)/bin/`.

jgwest:
See parent issue for details, this uses the same machinery for downloading a specific version of `controller-gen` to `(project path)/bin` that we are already using for kustomize.

Fixes #232

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ MKDOCS_RUN_ARGS?=
 CURRENT_DIR=$(shell pwd)
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
-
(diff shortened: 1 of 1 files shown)


wtam2018 (approved):
LGTM
@jgwest thanks!

Issue #192, opened 2021-04-08
Controller crash with git files generator having a number in the config.json


sboschman:
If you use a number / float64 in the `config.json` file for the git files generator it will crash / crash loop the controller.

Example from the docs, the number is quoted and treated as string:
```
{
  "asset_id": "11223344",
}
```

If someone does not quote the number it is treated as float64:
```
{
  "asset_id": 11223344,
}
```

Result:
```
E0407 18:12:34.409710       1 runtime.go:78] Observed a panic: &runtime.TypeAssertionError{_interface:(*runtime._type)(0x1c39b00), concrete:(*runtime._type)(0x1bb2d00), asserted:(*runtime._type)(0x1bf2080), missingMethod:""} (interface conversion: interface {} is float64, not string)

goroutine 328 [running]:

k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1c96280, 0xc00070fbc0)

	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/runtime/runtime.go:74 +0xa6

k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)

	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/runtime/runtime.go:48 +0x86

panic(0x1c96280, 0xc00070fbc0)

	/usr/local/go/src/runtime/panic.go:965 +0x1b9

github.com/argoproj-labs/applicationset/pkg/generators.(*GitGenerator).generateParamsFromGitFile(0xc00044e940, 0xc00057ca68, 0xc001653980, 0x1a, 0x1, 0xc00150ca90, 0x0, 0x1, 0x15)

	/workspace/pkg/generators/git.go:168 +0x6cd

github.com/argoproj-labs/applicationset/pkg/generators.(*GitGenerator).generateParamsForGitFiles(0xc00044e940, 0xc00057ca68, 0x0, 0x0, 0x0, 0x0, 0x0)

	/workspace/pkg/generators/git.go:124 +0x3c5

github.com/argoproj-labs/applicationset/pkg/generators.(*GitGenerator).GenerateParams(0xc00044e940, 0xc00057ca68, 0xc00057ca68, 0xc00114b260, 0x26, 0x0, 0x0)

	/workspace/pkg/generators/git.go:64 +0x65

[...]

PR #235, opened 2021-05-15
fix: handle conversion to string in git generator

chetan-rns:
Use `fmt.Sprintf` instead of type assertion for converting the incoming parameter to string.

Fixes: #192 

Signed-off-by: Chetan Banavikalmutt <chetanrns1997@gmail.com>


jgwest (approved):
Looks great, thanks @chetan-rns!

Issue #169, opened 2021-03-24
Add 'VERSION' file contents and latest commit ID to controller startup


jgwest:
For the purposes of problem determination and debugging, it would be beneficial if the controller pod logged the specific version of the ApplicationSet controller that was running. We can pull these values at build time from VERSION and git. 

See Argo CD's `common/version.go` and the Argo CD Makefile for an example of this.

PR #241, opened 2021-05-24
Log appset version during controller startup

chetan-rns:
Display the version of applicationset and git commit ID during controller startup log

Closes: #169

### DIFF
--- a/common/version.go
+++ b/common/version.go
@@ -1,0 +1,21 @@
+package common
+
+// version information populated during build time
+var (
+	version   = "" // value from VERSION file
+	gitCommit = "" // output of git rev-parse HEAD
+)
+
+// Version of the applicationset controller
+type Version struct {
+	Version   string
+	GitCommit string
+}
+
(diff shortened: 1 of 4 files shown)


common/version.go
```diff
@@ -0,0 +1,21 @@
+package common
+
+// version information populated during build time
+var (
+	version   = "v0.2.0" // value from VERSION file
```
jgwest, line 5: Can this be `version = "" // value from VERSION file`, or is there an advantage to having `v0.2.0` hardcoded in this file?

jgwest (changes requested):
Looks great! Just one question re: version string, then good to go.

common/version.go
chetan-rns: I thought we needed a default value here, but looks we can always get it from the VERSION file. Will change it to an empty string.

jgwest (approved):
Excellent, thanks again @chetan-rns!

PR #249, opened 2021-05-27
Enable the ability to set log level #238

### LINKED ISSUE #238: Support log-level for controller logs
**@jgwest EDIT:**
Enhancement Request: 
Add support for configurable log levels via [parameter](https://github.com/argoproj-labs/applicationset/blob/48a7b47da0552b9e5a518704e25489ab4a36645c/main.go#L81), eg the ability to set additional log levels beyond just the `--debug` parameter support we have now. For example, as requested, ability to set the log level to only show error or warn.

**Original Text from @dorroddorrod**
Hi,
Is there any way to control the log level for the application-set controller ?

rumstead:
closes #238

### DIFF
--- a/main.go
+++ b/main.go
@@ -185,7 +185,10 @@ func setLoggingLevel(debug bool, logLevel string) {
 		log.SetLevel(log.DebugLevel)
 	} else {
 		level, err := log.ParseLevel(logLevel)
-		setupLog.Error(err, "unable to parse loglevel", "loglevel", logLevel)
+		if err != nil {
+			setupLog.Error(err, "unable to parse loglevel", "loglevel", logLevel)
+			os.Exit(1)
+		}
 		log.SetLevel(level)
 	}
 }
--- a/main.go
+++ b/main.go
@@ -20,9 +20,10 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


jgwest:
@rumstead  Looks great to me, is it still WIP? Any other changes you are planning, or good to go?

rumstead:
G2G

main.go
```diff
@@ -177,3 +178,14 @@ func main() {
...
 	}
 }
+
+func setLoggingLevel(debug bool, logLevel string) {
+	// the debug flag takes precedence over the loglevel flag
+	if debug {
+		log.SetLevel(log.DebugLevel)
+	} else {
+		level, err := log.ParseLevel(logLevel)
+		setupLog.Error(err, "unable to parse loglevel", "loglevel", logLevel)
```
jgwest, line 188: Forgot that I had one question, @rumstead:

Should this be:
```
if err != nil {
  // log error about unable unable to parse loglevel?
} else {
  set level?
}
```

jgwest (changes requested):

main.go
rumstead: Yes, good catch! I committed as the wrong user and had a heck of a time undoing them.

jgwest (approved):
Looks great, thanks again @rumstead!

PR #253, opened 2021-05-31
Check path exists passed in application spec

mksha:
* Added error logic for path check
* updated test


mksha:
@jgwest  can you pls review it?

jgwest:
Hi @mksha, can you describe in the description the purpose of the change?

mksha:
TestGetFileContent is still failing on mac, bec of missing `/` in the expected string. Its better to just check the error kind not the actual content of error string.

replaced ioutil.ReadFile with os.Readfile() because it calls it under the hood.

mksha:
@jgwest  not sure github actions are failing , but same is passing on local.

DOCKER_PUSH=false make image

```
/Users/mohsharm/OpenSource/applicationset/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
/Users/mohsharm/OpenSource/applicationset/bin/controller-gen crd:crdVersions=v1 paths="./..." output:crd:artifacts:config=./manifests/crds/
KUSTOMIZE=/Users/mohsharm/OpenSource/applicationset/bin/kustomize hack/generate-manifests.sh
+++ dirname hack/generate-manifests.sh
++ CDPATH=
++ cd -- hack/..
++ pwd -P
+ SRCROOT=/Users/mohsharm/OpenSource/applicationset
+ AUTOGENMSG='# This is an auto-generated file. DO NOT EDIT'
+ KUSTOMIZE=/Users/mohsharm/OpenSource/applicationset/bin/kustomize
++ mktemp /tmp/appset-manifests.XXXXXX
+ TEMPFILE=/tmp/appset-manifests.NjWjQU
+ '[' '' '!=' '' ']'
+ IMAGE_NAME=argocd-applicationset
+ IMAGE_NAMESPACE=argoprojlabs
+ IMAGE_TAG=
+ '[' '' = '' ']'
++ git rev-parse --abbrev-ref HEAD
+ branch=Bugfix/temp-folder
+ [[ Bugfix/temp-folder = release-* ]]
+ '[' '' = '' ']'
+ IMAGE_TAG=latest
+ cd /Users/mohsharm/OpenSource/applicationset/manifests/base
+ /Users/mohsharm/OpenSource/applicationset/bin/kustomize edit set image argoprojlabs/argocd-applicationset=argoprojlabs/argocd-applicationset:latest
+ echo '# This is an auto-generated file. DO NOT EDIT'
+ cd /Users/mohsharm/OpenSource/applicationset/manifests/namespace-install
+ /Users/mohsharm/OpenSource/applicationset/bin/kustomize build .
+ mv /tmp/appset-manifests.NjWjQU /Users/mohsharm/OpenSource/applicationset/manifests/install.yaml
+ cd /Users/mohsharm/OpenSource/applicationset
+ chmod 644 manifests/install.yaml
+ /Users/mohsharm/OpenSource/applicationset/hack/verify-argo-cd-versions.sh
+ echo '# This is an auto-generated file. DO NOT EDIT'
+ cd /Users/mohsharm/OpenSource/applicationset/manifests/namespace-install-with-argo-cd
+ /Users/mohsharm/OpenSource/applicationset/bin/kustomize build .
+ mv /tmp/appset-manifests.NjWjQU /Users/mohsharm/OpenSource/applicationset/manifests/install-with-argo-cd.yaml
+ cd /Users/mohsharm/OpenSource/applicationset
+ chmod 644 manifests/install-with-argo-cd.yaml
go test -race -count=1 -coverprofile=coverage.out `go list ./... | grep -v 'test/e2e'`
?       github.com/argoproj-labs/applicationset [no test files]
?       github.com/argoproj-labs/applicationset/api/v1alpha1    [no test files]
ok      github.com/argoproj-labs/applicationset/pkg/controllers 4.264s  coverage: 67.6% of statements
ok      github.com/argoproj-labs/applicationset/pkg/generators  3.762s  coverage: 80.1% of statements
ok      github.com/argoproj-labs/applicationset/pkg/services    16.998s coverage: 79.4% of statements
ok      github.com/argoproj-labs/applicationset/pkg/services/scm_provider       14.299s coverage: 87.0% of statements
ok      github.com/argoproj-labs/applicationset/pkg/utils       4.842s  coverage: 71.2% of statements
docker build -t argoprojlabs/argocd-applicationset:latest .
```

pkg/services/repo_service.go
```diff
@@ -141,9 +140,9 @@ func (a *argoCDService) GetFileContent(ctx context.Context, repoURL string, revi
 		return nil, err
 	}
 
-	bytes, err := ioutil.ReadFile(filepath.Join(gitRepoClient.Root(), path))
+	bytes, err := os.ReadFile(filepath.Join(gitRepoClient.Root(), path))
 	if err != nil {
-		return nil, err
+		return nil, err.(*os.PathError).Err
```
jgwest, line 145: Can we return the error without casting it to (`*os.PathError`)? This seems like it could break us if the underlying implementation were to change. 

What if we just `return nil, err` here as before? Any issues there?

jgwest (changes requested):

pkg/services/repo_service.go
mksha: I dont think if that is true, its core lib. Its same as we are dealing with any logic. But i am fine with both. 

Let me replace it with err , but still we are comparing the error strings and it could change anytime so it will break the things.

So in both way, expectations are same. But i like the idea of checking the path inside that string so will just revert it back what it was with some fix in the test itself for correct path generation.

.golangci.yml
```diff
@@ -12,3 +12,9 @@ linters:
     - ineffassign
     - unconvert
     - unparam
+output:
+  format: github-actions
+issues:
+  exclude:
```
jgwest, line 18: @mksha can you describe what these are for?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+repos:
```
jgwest, line 1: Can you describe what this is doing?

jgwest (changes requested):

.golangci.yml
mksha: i saw in our github actions we are setting using cli arguments , but i think we can just set them in config file directly. @jgwest

.pre-commit-config.yaml
mksha: I have added a git pre-commit hook, that will easy the things for local development. So whenever i commit or push something it will run some things on local like go fmt, go mod tidy, lint, golangcilint and other and if everything is ok, then will allow me to commit something.

Its upto a particular person, if they enable the pre-commit hook on their local repo then this feature will be enabled otherwise it will be disabled and will not cause any issue for others.

.golangci.yml
jgwest: I see, thanks @mksha, can you remove the new excludes, I don't think those were present before. It seems like your PR still compiles without them.

.pre-commit-config.yaml
jgwest: I'd say lets keep these out of the repository for now, folks are free to maintain these locally if they wish.

.golangci.yml
mksha: @jgwest  i have not added anything new, its what we have in our ci https://github.com/argoproj-labs/applicationset/blob/master/.github/workflows/ci-build.yaml#L81

.pre-commit-config.yaml
mksha: I think it will help others also, because currently CI is not running on each commit. But will remove it for now.

.golangci.yml
jgwest: Ah, sorry about that @mksha, I forgot we had that and when I did `ripgrep` to find references, it ignores the `.github` folder :thinking:

jgwest (approved):
Looks great, thanks @mksha!

PR #264, opened 2021-06-13
docs: use image-push target to build and push an image

chetan-rns:
In the development guide it is recommended to use `make image` to build and push an image. `make image-push` would be the right target in this context.


jgwest (approved):
Agreed, thanks @chetan-rns

PR #267, opened 2021-06-15
Split generators into separate doc pages, plus general doc updates

### LINKED ISSUE #265: Split generators into separate doc pages, plus general doc updates
As the generator documentation grows, and the number of generators grow, it now makes sense to split the generators into their own RTD doc pages.

jgwest:
This PR:
- Moves the generators to their own pages
- Updates the table of contents for the above
- Tweaks to individual generator content
- Fixes the typo in the Getting Started filename, and adds a redirect for it.

Fixes #265


wtam2018 (approved):
LGTM
thanks @jgwest !

PR #282, opened 2021-07-02
Issue #170 is not solved when using server, only when using name for destination

### LINKED ISSUE #220: Issue #170 is not solved when using `server`, only when using `name` for destination
`ValidationDestination` is being used [here](https://github.com/argoproj-labs/applicationset/blob/v0.1.0/pkg/controllers/applicationset_controller.go#L560) to determine if we should remove the finalizer.  You can see in its implementation [here](https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go#L296), the entire function only does any validation when `dest.Name != ""`.  So when using `server` for the destination, the function passes and the finalizer is never removed.

jgwest:
This is the same fix as I made on Argo CD [here](https://github.com/argoproj/argo-cd/pull/6557) but since we support older Argo CDs in ApplicationSet controller, it needs to be made here as well.
- In addition to verifying using the name field (as before), we now also verify using the server field
- Unit test to verify the logic is correct

Fixes #220

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -445,7 +445,7 @@ func (r *ApplicationSetReconciler) removeFinalizerOnInvalidDestination(ctx conte
 		return nil
 	}
 
-	validDestination := true
+	var validDestination bool
 
 	// Detect if the destination is invalid (name doesn't correspond to a matching cluster)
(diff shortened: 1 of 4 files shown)


wtam2018 (approved):
LGTM
thanks @jgwest

PR #289, opened 2021-07-06
feat: add finalizers in application template

### LINKED ISSUE #288: Add option to use background deletion propagation policy for applications
Some resources in my applications only support background deletion. I couldn't find a way to set this option for applications created by an application set.

tenevdev:
This is my first contribution so excuse me if it's a wrong approach. I did it based on https://github.com/argoproj-labs/applicationset/pull/128, ran `make build` to generate the manifests, added a test by pattern-matching an existing one, ran `make test`. The interesting edits are in util.go and applcationset_types.go, I guess.

Fixes #288.


tenevdev:
I suspect I need to plug things together in this function but I'm a bit lost because I thought all metadata would be passed on transparently and not every field, also the fields don't really match the types so I guess this is a different version of the API? - https://github.com/argoproj-labs/applicationset/blob/d01eaf8d93081d0d1e2112f55fb70f11825aa214/pkg/controllers/applicationset_controller.go#L207

tenevdev:
> also the fields don't really match the types

TIL: promoted fields in golang 🤯

jgwest (approved):
Looks great, thanks @tenevdev! As part of the review process I added some unit tests to verify the expected behaviour, and everything worked as expected :+1: .

Issue #162, opened 2021-03-18
List generator should allow any key/value pair


moises-perez-tfs:
I find cluster `name` to be a more readable options than `url` for referencing clusters in an Application (or ApplicationSet).

For example:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
 name: guestbook
spec:
 generators:
 - list:
     elements:
     - cluster: engineering-dev
       url: REQUIRED_BUT_NOT_USED
     - cluster: engineering-prod
       url: REQUIRED_BUT_NOT_USED
 template:
   metadata:
     name: '{{cluster}}-guestbook'
   spec:
     project: default
     source:
       repoURL: https://github.com/argoproj-labs/applicationset.git
       targetRevision: HEAD
       path: examples/list-generator/guestbook/{{cluster}}
     destination:
       name: '{{cluster}}'
       namespace: guestbook
```

This example works fine when setting the appropriate cluster names but it requires us to set an unused `url` parameter for each element. This parameter is never used in the `template` above. I think using the cluster `name` in `destination` and avoiding URLs would be much easier to read.

If https://github.com/argoproj-labs/applicationset/issues/158 is ever implemented, that could be used as an alternative.

an-tex:
I'd even go as far as making none of the fields required. This way you could not only use the list generator to generate applications for different clusters, but for any kind (esp. useful if you only deploy to the local cluster only anyway). E.g. I'm just playing around with ApplicationSet to generate Applications for a subset of kustomize directories (as it's a subset you can't use the directories generator):
```scala
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: system-base
spec:
  generators:
    - list:
        elements:
          - values:
              name: elastic-operator
            cluster: ignored
            url: ignored
          - values:
              name: kube-state-metrics
            cluster: ignored
            url: ignored
          - values:
              name: openebs
            cluster: ignored
            url: ignored
          - values:
              name: telepresence
            cluster: ignored
            url: ignored
  template:
    metadata:
      name: '{{values.name}}'
    spec:
      source:
        path: base/system/{{values.name}}
```

The missing ApplicationSet template fields are populated by kustomization components depending on the environment

OmerKahani:
@jgwest what do you think about: 
[...]

PR #290, opened 2021-07-12
feat: allow any key/value pair in list generator

ishitasequeira:
Allow any key/value pair in list generator supporting backward compatibility for `values` field

Fixes: #162 

Signed-off : [ Ishita Sequeira](@ishitasequeira)


ishitasequeira:
@jgwest PTAL the PR is ready for review

pkg/generators/list.go
```diff
@@ -38,12 +38,28 @@ func (g *ListGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Appli
...
 		}
+
+		for key, value := range element {
+			if key == "values" {
+				values, ok := (value).(map[string]interface{})
+				if !ok {
+					return nil, fmt.Errorf("error parsing values map")
+				}
+				for k, v := range values {
+					params[fmt.Sprintf("values.%s", k)] = v.(string)
```
jgwest, line 56: It looks like this will panic if the user tries to include a non-string in the JSON, we should do a type check on this one as well, eg:
```go
value,  ok := v.(string)
if !ok { /* return error*/}
params[fmt.Sprintf("values.%s", k)] = value
```

pkg/generators/list.go
```diff
@@ -38,12 +38,28 @@ func (g *ListGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Appli
...
+				}
+			} else {
+				params[key] = value.(string)
```
jgwest, line 59: And we should type check (and return on error) for this one as well.

jgwest (changes requested):
Looks great, I'm really looking forward to this feature :smile: . I've added a few quick code tweak comments below, and one more thing: can you update the documentation under `docs/Generators-List.md`? Just need to change it to say that the user can now specify any key/value field they want (as long as they are `string`s), rather than just `cluster` and `url`.

I'm also going to do some quick manual testing to make sure it works as expected, I'll let you know if find any issues (but I'm not expecting any, the PR looks excellent)

jgwest (approved):
Looks great, thanks again @ishitasequeira!

PR #296, opened 2021-07-14
ApplicationSets that initially fail validation might not be re-reconciled

jgwest:
This PR:
- Now queues a reconcile on validation error
- Unit test to verify that behaviour
- Tweak to unrelated Git unit test

Fixes https://github.com/argoproj-labs/applicationset/issues/219


wtam2018 (approved):
LGTM
thanks @jgwest

PR #297, opened 2021-07-15
Document the policy parameter of the controller

### LINKED ISSUE #212: Document the policy parameter of the controller
The ApplicationSet controller has the ability to restrict how it applies cluster resource updates, for example, it can be set to create `Application` resources, but never delete them. 

We should document this feature and how to enable it.

See the `policy` param of `main.go` for details:
```
flag.StringVar(&policy, "policy", "sync", "Modify how application is synced between the generator and the cluster. Default is 'sync' (create & update & delete), options: 'create-only', 'create-update' (no deletion)")
```

jgwest:
Document the policy parameter of the controller, plus a more general writeup of how to control resource modification with ApplicationSet controller.

Fixes #212


docs/Controlling-Resource-Modification.md
```diff
@@ -0,0 +1,162 @@
...
+        # Insert new parameters here, for example:
+        # --policy create-only
+    # (...)
+```
+
+Save and exit the editor. Wait for a new `Pod` to start containing the updated parameters.
+
+### Or, B) Edit the `install.yaml` manifest for the ApplicationSet installation
+
+Rather than directly editting the cluster resource, you may instead choose to modify the installation YAML that is used to install the ApplicationSet controller:
```
wtam2018, line 95:
```suggestion
Rather than directly editing the cluster resource, you may instead choose to modify the installation YAML that is used to install the ApplicationSet controller:
```

wtam2018 (commented):
Spotted a typo.   Otherwise, LGTM.
thanks @jgwest

docs/Controlling-Resource-Modification.md
jgwest: Haha, I also often do this with the word `targeting` (too many t-s), thanks @wtam2018 :smile:

wtam2018 (approved):
LGTM
thanks @jgwest !

Issue #222, opened 2021-05-03
Support Webhooks trigger For Git generators in application set


mksha:
Currently in application set if we use git generator files, then it will poll the specified repo in every 3 min, but I think either we should allow the use to change the polling period or support the use of git webhook For Git generators so the changes to the files can be detected immediately rather than waiting for 3 min.

chetan-rns:
@jgwest I would like to work on this. Could you please assign it to me?

c0deaddict:
Would love to see this in applicationset :heart: Any update on this?

chetan-rns:
@c0deaddict  I've started looking into it. Will try to get the PR ready this week

c0deaddict:
Awesome! :heart:

PR #341, opened 2021-08-16
feat: refresh ApplicationSet resource with git generator using webhook

chetan-rns:
Currently, the repository used in git-generator will be polled every 3 mins. This commit adds support to refresh the application set using a git webhook. It exposes a service listening for incoming webhook payloads and triggers the application set controller.

Fixes: #222


pkg/utils/webhook.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (h *WebhookHandler) HandleEvent(payload interface{}) {
+	webURL, revision, touchedHead := getPayloadInfo(payload)
+	log.Infof("Received push event repo: %s, revision: %s, touchedHead: %v", webURL, revision, touchedHead)
+
+	appSetList := &v1alpha1.ApplicationSetList{}
+	err := h.client.List(context.Background(), appSetList, &client.ListOptions{})
+	if err != nil {
+		log.Warnf("Failed to list applicationsets: %v", err)
```
wtam2018, line 61: We probably should log it as an error.

pkg/utils/webhook.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+
+	urlObj, err := url.Parse(webURL)
+	if err != nil {
+		log.Warnf("Failed to parse repoURL '%s'", webURL)
```
wtam2018, line 67: We probably should log it as an error.

pkg/utils/webhook.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+	regexpStr := `(?i)(http://|https://|\w+@|ssh://(\w+@)?)` + urlObj.Hostname() + "(:[0-9]+|)[:/]" + urlObj.Path[1:] + "(\\.git)?"
+	repoRegexp, err := regexp.Compile(regexpStr)
+	if err != nil {
+		log.Warnf("Failed to compile regexp for repoURL '%s'", webURL)
```
wtam2018, line 73: We probably should log it as an error.

pkg/utils/webhook.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+
+	for _, appSet := range appSetList.Items {
+		for _, gen := range appSet.Spec.Generators {
+			// check if the ApplicationSet uses the git generator that is relevant to the payload
+			if gen.Git != nil && gitGeneratorUsesURL(gen.Git, revision, repoRegexp) && genRevisionHasChanged(gen.Git, revision, touchedHead) {
+				err := refreshApplicationSet(h.client, &appSet)
+				if err != nil {
+					log.Warnf("Failed to refresh ApplicationSet '%s' for controller reprocessing", appSet.Name)
```
wtam2018, line 83: We probably should log it as an error.

pkg/utils/webhook.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+func (h *WebhookHandler) Handler(w http.ResponseWriter, r *http.Request) {
+	var payload interface{}
+	var err error
+
+	switch {
+	case r.Header.Get("X-GitHub-Event") != "":
+		payload, err = h.github.Parse(r, github.PushEvent)
```
wtam2018, line 97: github/gitlab handler could be nil if their New() function have failed?

pkg/utils/webhook.go
```diff
@@ -0,0 +1,171 @@
...
+
+func NewWebhookHandler(namespace string, argocdSettingsMgr *argosettings.SettingsManager, client client.Client) *WebhookHandler {
+	// register the webhook secrets stored under "argocd-secret" for verifying incoming payloads
+	argocdSettings, err := argocdSettingsMgr.GetSettings()
+	if err != nil {
+		log.Errorf("Failed to get argocd settings: %v", err)
+	}
+	githubHandler, err := github.New(github.Options.Secret(argocdSettings.WebhookGitHubSecret))
+	if err != nil {
+		log.Warnf("Unable to init GitHub webhook: %v", err)
```
wtam2018, line 39: We don't log an error because github/gitlab secret is not required?

jgwest:
@chetan-rns If you rebase to master it should fix the E2E tests, FYI

pkg/utils/webhook.go
chetan-rns: I've changed it to log it as an error

pkg/utils/webhook.go
chetan-rns: I've updated the `NewGitHubHandler` to return an error. This function won't be called if there's an error while creating handler

jgwest (changes requested):
Content looks great! Rather than having webhook be its own page (like it is for Argo CD docs), let's move the contents of the new Webhook page into the existing `Generators-Git.md` file, adding it to the end of the document, eg:

```
## Git Generator: Files

(... previous content of Git Generator: Files...)

## Git Generator Webhook Configuration

When using a Git generator, ApplicationSet polls Git repositories (...)
```

(At first I thought we should have it be a separate page (like with Argo CD, and in your last commit) but I think it makes sense to include this content on the Git generator page, as it is applicable to both generator types, but is only applicable to these two generators. )

chetan-rns:
@jgwest I agree it makes more sense to have the webhook section in the existing Git Generator page. I've updated it. Please take a look :)

jgwest (approved):
Awesome, thanks @chetan-rns! Really great to get this feature in, which folks have been looking for :+1: .

main.go
```diff
@@ -197,3 +206,16 @@ func setLoggingLevel(debug bool, logLevel string) {
...
+	mux.HandleFunc("/api/webhook", webhookHandler.Handler)
+	go func() {
+		setupLog.Info("Starting webhook server")
+		err := http.ListenAndServe(":7000", mux)
+		if err != nil {
+			setupLog.Error(err, "failed to start webhook server")
+			os.Exit(1)
+		}
+	}()
+}
```
mkilchhofer, line 221: Hmm, is it a good idea to hardcode the port here as other port bindings are configurable: https://github.com/argoproj-labs/applicationset/blob/master/main.go#L84-L85
```golang
	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
	flag.StringVar(&probeBindAddr, "probe-addr", ":8081", "The address the probe endpoint binds to.")
```
Also it would be nice to make the webhook handler optional. This introduces a hard dependency on the ConfigMap `argocd-cm` which makes it really hard to find a good solution for the helm chart update to 0.3.0:
https://github.com/argoproj/argo-helm/pull/1070

chetan-rns: Hi @mkilchhofer, thanks for bringing this up. I agree the webhook address should be configurable. I've raised a PR for this https://github.com/argoproj-labs/applicationset/pull/450. It also makes the webhook handler optional by removing the hard exit.

mkilchhofer: Cool. Also worked on this :-D but you were faster, thanks. 🙏

Issue #354, opened 2021-08-26
The git files generator is slow


alexmt:
# Summary

Git files generator is consuming too much time because it execute `git fetch` `git checkout` for each discovered file.

# Steps

* Follow steps from [documentation](https://github.com/argoproj-labs/applicationset/blob/master/docs/Generators-Git.md#git-generator-files) to configure file generator
* Make sure repository has ~30 files
* Check controller logs: controller executes  `git fetch` `git checkout` for each discovered file.

# Expected behavior

Controller should run  `git fetch` `git checkout` once per repo

PR #355, opened 2021-08-26
fix: improve git files generator performance

alexmt:

Closes https://github.com/argoproj-labs/applicationset/issues/354

PR changes repo services implementation: replaces `GetFilePaths` and `GetFileContent` methods with `GetFiles` that checkout/fetch repo once and return content of discovered files.


jgwest (approved):
Looks great, thanks @alexmt!  The only minor disadvantage to this approach is it will read the full file contents of matching paths into memory (causing spiky heap allocation)  but since it will only read files that match the path (which tend to be smallish `.json` and `.yaml` files), this is perfectly fine.

PR #357, opened 2021-08-26
docs: fix includeSubgroups reference name

yogeek:


jgwest (approved):
Oops :sweat_smile: , good catch, thanks @yogeek!

PR #370, opened 2021-09-15
feat: report error conditions/status for ApplicationSet

### LINKED ISSUE #71: ApplicationSet should report error conditions/status based on the result of generator/template resuls

ishitasequeira:
Report error conditions/status fo ApplicationSet 

Fixes #71 

Signed off by [Ishita Sequeira](@ishitasequeira )


manifests/crds/argoproj.io_applicationset.yaml
```diff
@@ -0,0 +1,6414 @@
+
+---
```
jgwest, line 2: Think this file can be removed? In this version of the file, the plural is `applicationset` rather than `applicationsets` for some reason :thinking:.

jgwest (commented):
So far so good, thanks @ishitasequeira! Some updates requested to the condition logic, so that we don't update the `status` field too often, and so that we always have all 3 conditions present. Take a look and let me know what you think.

ishitasequeira:
Thanks @jgwest for the review. I have made the requested changes, PTAL.

alexmt:
Looking forward to this PR! Will contribute custom health check for app set as soon as this PR is merged

pkg/controllers/applicationset_controller_test.go
```diff
@@ -1732,6 +1752,16 @@ func TestValidateGeneratedApplications(t *testing.T) {
 					assert.True(t, foundMatch, "Unble to locate expected error: %s", cc.expectedErrors)
 					matched = matched || foundMatch
 				}
+				// validation message was returned: it should be expected
+				matched = false
```
jgwest, line 1756: I think this `matched = false` should be removed, otherwise the `match` logic above this line is ignored (correct me if I am wrong)

jgwest (changes requested):
Looks great @ishitasequeira, one tiny test tweak than good to go!

pkg/controllers/applicationset_controller_test.go
ishitasequeira: I guess I missed out an assert statement before this line. I have added that. The above `matched=false` will reset the value for matched for test in the section below it.

jgwest (approved):
Looks great, thanks @ishitasequeira! It's great to finally have error reporting as part of ApplicationSet CR :tada:.

PR #372, opened 2021-09-16
feat: Gracefully handle application validation errors

### LINKED ISSUE #356: Gracefully handle application validation errors
# Summary

Application set validates generated applications (which is good) but refuses to create any application even if only one of the generates apps is invalid. Instead, the controller should generate valid applications and skip only invalid ones.

Here is a sample use case where current behavior causes confusion. 
* App set generates an application that references cluster A
* Generator source has changes so now app set produces 2 applications and Cluster A got removed from Argo CD.
* Controller detects the error in the first A (since cluster no longer registered) 
* User assuming that something is wrong with new application and does not know how to fix the issue

# Proposal 

Two changes are proposed:

* Application Set controller should generate valid applications and skip only invalid once
* The applications generation error should be reported in `status.conditions` field of the ApplicationSet CR:

```yaml
status:
  conditions:
    type: ApplicationsGenerated
    status: False
    reason: ApplicationValidationFailed
    message: Cluster A is not registered
```

alexmt:

Closes #356 

PR changes how the ApplicationSet controller handles ApplicationSets that generate invalid apps. Instead of refusing to handle such app sets controller will process valid apps and log information about invalid apps.

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -96,7 +96,8 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
-	if validateError := r.validateGeneratedApplications(ctx, desiredApplications, applicationSetInfo, req.Namespace); validateError != nil {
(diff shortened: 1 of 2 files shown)


jgwest (approved):
LGTM, thanks @alexmt!

PR #373, opened 2021-09-17
feat: support `-logformat=json`

### LINKED ISSUE #57: Support JSON-format logs
Argo CD uses flag `--logformat=json|text` with `text` as default.

shivjm:
Also forces controller-runtime to produce pure JSON. Modelled after (but not identical to) ArgoCD:

https://github.com/argoproj/argo-cd/blob/e44fa434abcfd622f13cf8a0107bc84752184f36/util/log/logrus.go#L35-L51

Closes #57

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jgwest (approved):
LGTM, thanks @shivjm!

PR #380, opened 2021-09-29
Skip E2E tests that require GitHub token, if not specified

jgwest:
This PR adds a new function `IsGitHubSkippedTest`, which returns true if the test should be skipped because it requires a GitHub API Token and  one has not been provided.

Unfortunately, GitHub Actions cannot use repository secrets, so we need to skip these tests for PRs. These tests will still run on the `main` branch after a PR has merged, which will alert us of any issues.


wtam2018 (approved):
LGTM
Thanks @jgwest !

Issue #364, opened 2021-09-02
Clusters with underscore in name causing problem with variable templating


r0bj:
Clusters with underscore in name causing problem with variable templating.

Example.
```
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: myapp
spec:
  generators:
  - clusters: {}
  template:
    metadata:
      name: 'myapp-{{name}}'
    spec:
      project: default
      source:
        path: apps/myapp
        repoURL: git@github.com:myorg/apps.git
        targetRevision: HEAD
      destination:
        name: '{{name}}'
        namespace: prod
```

Everything work fine unless cluster name consist underscore, e.g. `gke_myproject_us-west1-a_cluster1` (similar to how GKE clusters are named). In this case applicationset controller generates error message:
```
time="2021-09-02T21:09:13Z" level=error msg="error occurred during application generation: application destination spec is invalid: unable to find destination server: there are no clusters with this name: gke-myproject-us-west1-a-cluster1"
```

So it seems that applicationset controller replaced "_" to "-" in cluster name but now variable {{name}} is somehow broken, it is no longer store valid cluster name. So using that variable as a destination cluster is no longer works.

Maybe instead of modifying variable {{name}} by controller we can add new variable e.g. {{nameNormalized}} and leave {{name}} unchanged? It would give more transparency and predictability to the process.

jgwest:
Agreed, it doesn't make sense to sanitize the `name` field, if the corresponding name can't be used in `.destination.name` field :thinking: .

:+1:  for `{{nameNormalized}}` as being the normalized version of the name field, and reverting the `name` field back to its original behaviour of providing the unmodified version of the string. 

(This would have backwards compatibility implications for some users, so we would need to call it out in our upgrade notes.)

PR #390, opened 2021-10-15
fix: new variable for the normalized version of the name field

chetan-rns:
The cluster name field in the destination expects the original cluster name instead of a sanitized version. Introduce a new variable `{{nameNormalized}}`  for the sanitized version and retain the original cluster name in the `{{name}}` field  

Closes: #364


jgwest (approved):
LGTM, thanks @chetan-rns !

PR #403, opened 2021-11-02
remove namespace from Deployment in index.yaml

ohauer:
Remove namespace from Deployment.

I was searching why I got an issue with the following patch and it was because of the namespace in kind Deployment
Looking into the install.yaml of argo-cd and others the difference is the namespace in kind: Deployment in index.yaml.

Two possible pathes:
1. remove the namespace from the Deployment (it is normally given/ overwritten with kustomize)
2. force users to add additional namespace: to the patch (else kustomize will fail and complain with the following message
```
no matches for Id apps_v1_Deployment|~X|argocd-applicationset-controller; failed to find unique target for patch apps_v1_Deployment|argocd-applicationset-controller"
```

```sh
cat > overlay/clusterX/kustomization.yaml <<EOM
kind: Kustomization
namespace: argocd
resources:
- ../../base/

patchesStrategicMerge:
- patch-limits.yaml
EOM

cat > overlay/clusterX/patch-limits.yaml << EOM
apiVersion: apps/v1
kind: Deployment
metadata:
  name: argocd-applicationset-controller
spec:
  template:
    spec:
      containers:
      - name: argocd-applicationset-controller
        resources:
          requests:
            cpu: 50m
            memory: 64Mi
EOM
```

### DIFF
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -5415,7 +5415,6 @@ metadata:
     app.kubernetes.io/name: argocd-applicationset-controller
     app.kubernetes.io/part-of: argocd-applicationset
   name: argocd-applicationset-controller
-  namespace: argocd
 spec:
   selector:
     matchLabels:


jgwest:
@ohauer The file you are proposing to change in this PR is auto-generated by `make manifests`, using kustomize, and should not modified directly. 

Note that Argo CD's `install.yaml` also hardcodes a namspace: https://github.com/argoproj/argo-cd/blob/0d4e40ed6a7c5f3854d47f11f30b6c0606548601/manifests/install.yaml#L2821

I'm not sure what you're trying to accomplish: are you trying to run Argo CD in a different namespace?

ohauer:
I found this by setting resource limits/requests to the kustomize based deployments. for argocd the sample patch from initial request works but not for the applicationcontroller.
For applicationcontroller the patch requires additional the namespace in the patch else kustomize will complain the patch is not spacific enough.

You are right, argo-cd also uses namespace but **only in ClusterRoleBinding and in context to the ServiceAccount** (Line 2821 and 2838) and nowhere else.

jgwest (changes requested):
Not sure this is the right approach, see above.

jgwest:
Fixed by https://github.com/argoproj/applicationset/pull/474

Issue #344, opened 2021-08-19
Support 'union' generator


alexmt:
# Summary 

Support generator that allows to "union" several generators and override output of one generator with values of another generator.

The new generator is supposed to solve the following use cases: 
* User wants to install the same application into all clusters with default parameters but override some parameters in couple clusters
* User wants to install the same application into all clusters but split clusters into "waves". Each wave will have couple of different parameters

# Proposal

Introduce `union` generator that allows specifying list of sub-generators and list of merge keys. The new generator should unite output of merge generators. The merge keys should be taken into account to detect duplicates and override values:

Example:

```
 generators:
  - union:
      mergeKeys: # specifies which keys should be used to detect duplicates and apply overrides
      - url 
      generators:
        - clusters:
            selector:
              matchLabels:
                argocd.argoproj.io/secret-type: cluster
              matchExpressions:
                - {key: cluster.name, operator: NotIn, values: [excluded-cluster]} # matchExpressions should allow to exclude clusters
            values: # default values for all clusters
              install-kafka: true

        - items:
            elements:
              cluster: engineering-dev
            values: # overrides for dev
              install-kafka: false

```

rumstead:
#323 feels related here

rumstead:
@jgwest how we feeling about this? I know @alexmt's history with ArgoCD :). I feel like this issue combined with #362 will accomplish #323 thus has me very interested haha.

crenshaw-dev:
My understanding is that generated param maps should be merged if and only if the values for all keys in `mergeKeys` are the same. For example:

```
mergeKeys: a, b
P_1 = {a: 1, b: 1, c: 1}
P_2 = {a: 1, b: 2, c: 2}
P_3 = {a: 1, b: 1, c: 3}
// P_1 and P_3 are merged (with P_3 values taking priority) because _both_ a and b values match. P_2 is not merged.
union(P_1, P_2, P_3) == [{a: 1, b: 1, c: 3}, {a: 2, b: 2, c: 3}]
```

Is that correct?

PR #404, opened 2021-11-02
feat: support merge generator

crenshaw-dev:
fixes https://github.com/argoproj-labs/applicationset/issues/344


api/v1alpha1/applicationset_types.go
```diff
@@ -93,16 +110,85 @@ type ApplicationSetBaseGenerator struct {
...
+}
+
+// ToMatrixGenerator converts a NestedMatrixGenerator to a MatrixGenerator. This conversion is for convenience, allowing
+// a NestedMatrixGenerator to be used where a MatrixGenerator is expected (of course, the converted generator will have
+// no override template).
+func (g NestedMatrixGenerator) ToMatrixGenerator() *MatrixGenerator {
+	return &MatrixGenerator{
+		Generators: g.Generators.toApplicationSetNestedGenerators(),
+	}
+}
```
crenshaw-dev, line 160: I'm not sure I love this pattern. Would it be better to use reflection, or have `MatrixGenerator` and `NestedMatrixGenerator` implement the same interface?

Option 1 means reflection (lose some type checking at compile time).

Option 2 means defining redundant getters.

Meh, I don't love the status quo or either of the alternatives.

main.go
```diff
@@ -157,18 +157,30 @@ func main() {
...
-		os.Exit(1)
+	topLevelGenerators := map[string]generators.Generator{
+		"List":                    terminalGenerators["List"],
+		"Clusters":                terminalGenerators["Clusters"],
+		"Git":                     terminalGenerators["Git"],
+		"SCMProvider":             terminalGenerators["SCMProvider"],
+		"ClusterDecisionResource": terminalGenerators["ClusterDecisionResource"],
+		"PullRequest":             terminalGenerators["PullRequest"],
+		"Matrix":                  generators.NewMatrixGenerator(nestedGenerators),
+		"Union":                   generators.NewUnionGenerator(nestedGenerators),
```
crenshaw-dev, line 179: These annoy me because of the maintenance overhead (when a new generator is added, you have to remember to add it in all three places). But I can't think of a better alternative.

pkg/generators/git.go
```diff
@@ -48,8 +48,8 @@ func (g *GitGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Applic
 		return nil, EmptyAppSetGeneratorError
 	}
 
-	if appSetGenerator.Git == nil {
-		return nil, EmptyAppSetGeneratorError
+	if appSetGenerator.ApplicationSetTerminalGenerator == nil || appSetGenerator.ApplicationSetTerminalGenerator.Git == nil {
+		return nil, nil
```
crenshaw-dev, line 52: These checks are inconsistent. Some return `EmptyAppSetGeneratorError` and others return `nil`. I've standardized on `nil` for no particular reason.

These lines should never be hit. As far as I can tell they're just safety checks because the types technically allow this condition.

pkg/generators/pull_request.go
```diff
@@ -51,8 +51,8 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 		return nil, EmptyAppSetGeneratorError
 	}
 
-	if appSetGenerator.PullRequest == nil {
-		return nil, EmptyAppSetGeneratorError
+	if appSetGenerator.ApplicationSetTerminalGenerator == nil || appSetGenerator.ApplicationSetTerminalGenerator.PullRequest == nil {
+		return nil, nil
```
crenshaw-dev, line 55: These checks are inconsistent. Some return `EmptyAppSetGeneratorError` and others return `nil`. I've standardized on `nil` for no particular reason.

These lines should never be hit. As far as I can tell they're just safety checks because the types technically allow this condition.

pkg/generators/scm_provider.go
```diff
@@ -48,8 +48,8 @@ func (g *SCMProviderGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 		return nil, EmptyAppSetGeneratorError
 	}
 
-	if appSetGenerator.SCMProvider == nil {
-		return nil, EmptyAppSetGeneratorError
+	if appSetGenerator.ApplicationSetTerminalGenerator == nil || appSetGenerator.ApplicationSetTerminalGenerator.SCMProvider == nil {
+		return nil, nil
```
crenshaw-dev, line 52: These checks are inconsistent. Some return `EmptyAppSetGeneratorError` and others return `nil`. I've standardized on `nil` for no particular reason.

These lines should never be hit. As far as I can tell they're just safety checks because the types technically allow this condition.

crenshaw-dev:
@alexmt @jgwest can y'all take a look at this?

.github/workflows/ci-build.yaml
```diff
@@ -252,7 +252,7 @@ jobs:
         run: |
           set -x
           cd "$GITHUB_WORKSPACE/applicationsets"
-          kubectl apply -f manifests/crds/argoproj.io_applicationsets.yaml
+          kubectl create -f manifests/crds/argoproj.io_applicationsets.yaml
```
leoluz, line 255: This will probably fail if the build image already has this CRD installed. Is it guaranteed that we are always getting a fresh image?

api/v1alpha1/applicationset_types.go
```diff
@@ -70,21 +70,38 @@ type ApplicationSetTemplateMeta struct {
...
+// MergeGenerator).
+type ApplicationSetNestedGenerator struct {
+	List                    *ListGenerator         `json:"list,omitempty"`
+	Clusters                *ClusterGenerator      `json:"clusters,omitempty"`
+	Git                     *GitGenerator          `json:"git,omitempty"`
+	SCMProvider             *SCMProviderGenerator  `json:"scmProvider,omitempty"`
+	ClusterDecisionResource *DuckTypeGenerator     `json:"clusterDecisionResource,omitempty"`
+	PullRequest             *PullRequestGenerator  `json:"pullRequest,omitempty"`
+	Matrix                  *NestedMatrixGenerator `json:"matrix,omitempty"`
+	Merge                   *NestedMergeGenerator  `json:"merge,omitempty"`
```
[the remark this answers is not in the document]
leoluz: Im sorry but I don't understand why embedded structs makes it easier to miss nil checks. I believe it is the same.

api/v1alpha1/applicationset_types.go
leoluz: Personally I would go with defining a common interface which brings other advantages like more flexible code and easier to test/mock. My second choice would be having the conversion functions like the way you implemented which seems fine to me as well.
The solution I would really avoid is going with reflection as this is such a simple problem to handle and there are many negative side effects in going with reflection like:
- lower maintainability
- higher complexity
- harder to test
- losing compile time type checking like you described

pkg/generators/generator_spec_processor_test.go
```diff
@@ -0,0 +1,57 @@
+package generators
```
leoluz, line 1: It seems you are just using exported symbols having the test validation from the consumer perspective (blackbox test). This is really good type of test to make sure you are exporting just the necessary symbols. In this case you can define the package name with the `_test` suffix (`generators_test`) which is a Go builtin functionality. More on this here:
https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language

pkg/generators/git.go
leoluz: Actually I would do the opposite as returning nil in the second parameter could lead in a possible `nil pointer error` for consumers of this function eventually. I would standardize and return `EmptyAppSetGeneratorError` everywhere when this condition is met.

pkg/generators/matrix_test.go
```diff
@@ -95,14 +122,14 @@ func TestMatrixGenerate(t *testing.T) {
 		},
 	}
 
-	for _, c := range testCases {
-		cc := c
+	for _, testCase := range testCases {
+		testCaseCopy := testCase // Since tests may run in parallel
```
leoluz, line 126: Test cases won't run in parallel unless you call `t.Parallel()` explicitly from your test code which is something that I always do to make sure the tests executed faster.

pkg/generators/merge.go
```diff
@@ -0,0 +1,192 @@
+package generators
```
leoluz, line 1: Missing godocs in this file

pkg/generators/merge.go
```diff
@@ -0,0 +1,192 @@
...
+
+	deDuplicatedMergeKeys := make(map[string]bool, len(mergeKeys))
+	for _, mergeKey := range mergeKeys {
+		deDuplicatedMergeKeys[mergeKey] = false
+	}
+
+	paramSetsByMergeKey := make(map[string]map[string]string, len(paramSets))
+	for _, paramSet := range paramSets {
+		paramSetKey := make(map[string]string)
+		for mergeKey, _ := range deDuplicatedMergeKeys {
```
leoluz, line 103: Go allows you omit the second value if you are just interested in the first one:
This should work the same way:
```Go
for mergeKey := range deDuplicatedMergeKeys {
```

pkg/generators/merge_test.go
```diff
@@ -0,0 +1,309 @@
...
+			mergeKeys: []string{"a", "b"},
+			expected: []map[string]string{
+				{"a": "1", "b": "3", "c": "added", "d": "added"},
+				{"a": "2", "b": "2"},
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		testCaseCopy := testCase // since tests may run in parallel
```
leoluz, line 150: Please add `t.Parallel()` to hint go test to run the test cases in parallel

pkg/generators/pull_request.go
leoluz: I would revert and return `EmptyAppSetGeneratorError` everywhere when this condition is met.

leoluz (changes requested):

pkg/generators/scm_provider.go
leoluz: same here:
https://github.com/argoproj-labs/applicationset/pull/404#discussion_r746707169

api/v1alpha1/applicationset_types.go
crenshaw-dev: I need to take a few hours one day and attempt the interface solution to see how it feels.

I think the good news is that it's completely an implementation detail which could be applied later. (Let me know if that assumption might be incorrect. If that's the case, we can evaluate the interface approach before merging this.)

pkg/generators/git.go
crenshaw-dev: That seems completely reasonable. Will do.

pkg/generators/generator_spec_processor_test.go
crenshaw-dev: Ah nice, that's precisely what I searched for and somehow missed.

pkg/generators/matrix_test.go
crenshaw-dev: Will change to parallel then!

pkg/generators/merge.go
crenshaw-dev: Aw c'mon docs are for nerds.

jk, will add. :-)

api/v1alpha1/applicationset_types.go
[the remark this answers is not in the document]
crenshaw-dev: Suppose I have this code:

```go
type Book struct {
	Title string
}

func (b *Book) GetLowerCaseTitle() string {
	return toLowerCase(b.Title)
}
```

Then I decide to refactor to use an embedded struct:

```go
type BookMetadata struct {
	Title string
}

type Book struct {
	BookMetadata
}

func (b *Book) GetLowerCaseTitle() string {
	// this still compiles, because `b.Title` is syntactic sugar for `b.BookMetadata.Title`
	return toLowerCase(b.Title)
}

func createBook() Book {
	return Book{
		BookMetadata: nil,  // this compiles
	}
}
```

After the refactor, I have a bunch of references to now-embedded fields which _look_ immune to nil reference errors, but can still throw nil reference errors.

It would seem that embedding structs isn't like a macro (simply expanding the field names and literally "embedding" them in the containing struct). It's just a bit of syntactic sugar for composition, plus some nice default behavior from the json marshaller.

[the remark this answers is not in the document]
crenshaw-dev: ^ that's wrong. After the refactor, `Title` _cannot_ be nil. In my previous attempt, I'd used a pointer to an embedded struct, which could be nil. I'm not sure why I went with the pointer. Time permitting, I'll try with the non-pointer embedded struct.

.github/workflows/ci-build.yaml
crenshaw-dev: Looks to me as if it always spins up a fresh k3s before running the tests.

crenshaw-dev:
@leoluz I think I've made all the main changes. I've left a couple "optional" changes to look into if I have time.

pkg/generators/git_test.go
```diff
@@ -141,12 +141,15 @@ func TestGitGenerateParamsFromDirectories(t *testing.T) {
...
+	for _, testCase := range cases {
+		testCaseCopy := testCase
+
+		t.Run(testCaseCopy.name, func(t *testing.T) {
+			t.Parallel()
+
 			argoCDServiceMock := argoCDServiceMock{mock: &mock.Mock{}}
 
-			argoCDServiceMock.mock.On("GetDirectories", mock.Anything, mock.Anything, mock.Anything).Return(c.repoApps, c.repoError)
+			argoCDServiceMock.mock.On("GetDirectories", mock.Anything, mock.Anything, mock.Anything).Return(testCase.repoApps, testCase.repoError)
```
leoluz, line 152: make sure you only reference `testCaseCopy` inside this `t.Run` and not `testCase` or you will have unpredictable results running it in parallel

leoluz (changes requested):

leoluz (approved):
LGTM

docs/Generators-Merge.md
```diff
@@ -0,0 +1,144 @@
+# Merge Generator
+
+The merge generator combines parameters produced by the base (first) generator with matching parameter sets produced by other generators. A _matching_ parameter set has the same values for the configured _merge keys_. _Non-matching_ parameter sets are discarded. Override precedence is bottom-to-top. The values from a matching parameter set produced by generator 3 will take precedence over the values from the corresponding parameter set produced by generator 2.
```
jgwest, line 3: Tiny tweak to the first sentence

> The merge generator combines parameters produced by the base (first) generator with matching parameter sets produced by other generators.

To

"The merge generator combines parameters produced by the base (first) generator with matching parameter sets produced by **subsequent** generators."

pkg/generators/merge.go
```diff
@@ -0,0 +1,201 @@
...
+		for mergeKey := range deDuplicatedMergeKeys {
+			paramSetKey[mergeKey] = paramSet[mergeKey]
+		}
+		paramSetKeyJson, err := json.Marshal(paramSetKey)
+		if err != nil {
+			return nil, err
+		}
+		paramSetKeyString := string(paramSetKeyJson)
+		if _, exists := paramSetsByMergeKey[paramSetKeyString]; exists {
+			return nil, NonUniqueParamSets
```
jgwest, line 119: It couldn't hurt to report the specific `paramSetKeyString` that is causing the problem, in the error text, to help the user figure out where the problem is.

Makefile
```diff
@@ -84,7 +84,7 @@ start-e2e: # Ensure the PlacementDecision CRD is present for the ClusterDecision
 	kubectl apply -f https://raw.githubusercontent.com/open-cluster-management/api/a6845f2ebcb186ec26b832f60c988537a58f3859/cluster/v1alpha1/0000_04_clusters.open-cluster-management.io_placementdecisions.crd.yaml
 	NAMESPACE=argocd-e2e "dist/argocd-applicationset" --metrics-addr=:12345 --probe-addr=:12346 --argocd-repo-server=localhost:8081 --namespace=argocd-e2e
 
-# Begin the tests, targetting the standalone controller (started by make start-e2e) and the e2e argo-cd (started by make start-e2e)
+# Begin the tests, targeting the standalone controller (started by make start-e2e) and the e2e argo-cd (started by make start-e2e)
```
jgwest, line 87: Thanks, one of these days I will learn to correctly spell targeting, but not today :upside_down_face:.

jgwest (changes requested):
Looks great @crenshaw-dev! Really nice, detailed work :tada:.

Just a couple of requests:
- It would be good to get an E2E test for the merge generator, as we have for the other generators. It should be quick to put it together if you base it on the `matrix_e2e_test.go`
- Plus some quick tweaks mentioned inline in the review.

(As you said below, the need for nesting makes things a bit convoluted, but that's a necessary evil due to CRD's lack of support for recursion, and our desire to continue to support CRD validation and type definitions.)

pkg/utils/map_test.go
```diff
@@ -25,30 +25,32 @@ func TestCombineStringMaps(t *testing.T) {
...
+			right:       map[string]string{"a": "b", "testCase": "d"},
 			expected:    map[string]string{"a": "b", "foo": "bar"},
 			expectedErr: errors.New("found duplicate key a with different value, a: fail ,b: b"),
 		},
 		{
 			name:        "pass if keys & values are the same",
 			left:        map[string]string{"foo": "bar", "a": "b"},
-			right:       map[string]string{"a": "b", "c": "d"},
-			expected:    map[string]string{"a": "b", "c": "d", "foo": "bar"},
+			right:       map[string]string{"a": "b", "testCase": "d"},
```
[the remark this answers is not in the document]
crenshaw-dev: lol nice, good catch

Makefile
crenshaw-dev: For me it's one of those words that I just spell twice and let spellcheck figure it out. 😆

crenshaw-dev:
I'm confused by the build failure. The field which is missing from the committed manifests was recently added (https://github.com/argoproj/argo-cd/pull/7249). But as far as I can tell, it wasn't added to the argo-cd version in go.sum (2.1.0).

Even weirder (to me) is that the logs don't mention downloading argo-cd at all. Guess I need to learn more about go modules.

jgwest:
@crenshaw-dev Don't worry about it, I think what you have is good, I'll take a look and fix it as needed

crenshaw-dev:
@jgwest cool, thanks. I still need to add the e2e test, which I should be able to do this afternoon.

jgwest:
@crenshaw-dev - any luck with this, and/or any issues that you are hitting that I can help with? I'm working on the ApplicationSet v0.3.0 release, and it would be good to get this in if you think that's possible, but no pressure :smile:.

crenshaw-dev:
@jgwest I've been unable to get my e2e test env. working. I'll start a Slack thread. I'm sure I'm missing something simple. :-)

jgwest:
@crenshaw-dev  One other issue I've noticed is that the ApplicationSet controller install YAML can no longer be applied; I see now why you changed the install script from `kubectl create` to `kubectl apply`.., with the error being "The CustomResourceDefinition "applicationsets.argoproj.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes"

crenshaw-dev:
@jgwest yep... I'll double-check that there's no documentation which requires `apply` instead of `create`.

crenshaw-dev:
@jgwest added e2e tests. You'll probably also want to glance at the docs commit.

crenshaw-dev:
Looks like the doc lint error is a cert issue on kustomize.io. My guess is they'll fix the cert soon. Probably safe to ignore the error.

jgwest:
@crenshaw-dev  -

Switching from `kubectl apply` to `kubectl create` has a few major issues:
- It makes upgrade more difficult: we would need to ask the user to `kubectl delete && kubectl create` their application set install, vs the current model which is an in-place upgrade with apply. The in-place apply is how both [applicationset controller](https://argocd-applicationset.readthedocs.io/en/stable/Getting-Started/#a-install-applicationset-into-an-existing-argo-cd-install) and [Argo CD](https://argo-cd.readthedocs.io/en/stable/getting_started/#1-install-argo-cd) work now.
    - I'm not even sure if it's possible to delete a CRD if there are CRs that are still using it? IIRC I've seen mixed results here in the past. It means, to upgrade, you might need to delete all the ApplicationSet CRs as well.
- As you are aware, we are [working on merging ApplicationSet controller into Argo CD](https://github.com/argoproj/argo-cd/issues/7351). So this change would also affect Argo CD, because the ApplicationSet CR would now be a part of the Argo CD  install yaml. Thus we would require Argo CD users to `kubectl delete && kubectl create` upgrade Argo CD using the create/delete approach as well.

@alexmt have y'all hit this with Argo CD CR, and/or any suggestions? This would unfortunately affect Argo CD as wel,l if we were to [combine argo cd/appset install YAMLs](https://github.com/argoproj/argo-cd/issues/7351).

crenshaw-dev:
I'm a bit ignorant of the subtleties here... Couldn't a user perform a non-disruptive upgrade using `kubectl replace`?

jgwest (approved):
LGTM, thanks again @crenshaw-dev, great work.

(And thanks @leoluz for reviewing!)

Issue #407, opened 2021-11-04
Question about path parameter in Git Generator: Files


mbarecki:
Hello,

Let me start with saying that Argo CD ApplicationSet Controller is great.

I would like to ask question about  `Git Generator: Files` and `{{path}}` parameter.

In [Generators-Git.md](https://github.com/argoproj-labs/applicationset/blob/master/docs/Generators-Git.md#git-generator-files) document description of `{{path}}` is:
```
{{path}}: The path to the matching configuration file within the Git repository. Example: /cluster-config/staging/config.json
```

When looking at go test [TestGitGenerateParamsFromFiles](https://github.com/argoproj-labs/applicationset/blob/4f72703061f59c6b9b93ace5b8428bce22cb2891/pkg/generators/git_test.go#L232), file name is not included in expected value.

Is example value `/cluster-config/staging/config.json` returned by `{{path}}` from documentation not correct by containing file name in it ?

jgwest:
Good catch! Yes, the `Generators-Git.md` example is wrong. The correct behaviour is described in the original issue: https://github.com/argoproj-labs/applicationset/issues/258#issuecomment-856140869

PR #408, opened 2021-11-07
Fix Git Generator Files path example in docs (#407)

mbarecki:
Fix path parameter example in `Generators-Git.md`.

Relates to #407.


jgwest (approved):
LGTM, thanks @mbarecki!

PR #366, opened 2021-09-07
Feat/pull request generator

shmurata:
The new generator proposed at https://github.com/argoproj-labs/applicationset/issues/210 has been implemented as a Pull Request Generator.
Although the Pull Request Generator is similar to the SCM Provider Generator in some aspects, it seems that various changes are necessary to incorporate it into the SCM Provider Generator, so I have implemented it as a separate generator.
In the future, it may be possible to merge them.

I've implemented it to add other providers as well, but this PR only supports Github.
I can't implement other providers because I don't use them.

The requeue interval is polled in this PR, but I'd like to add webhook support to this generator after https://github.com/argoproj-labs/applicationset/pull/341 is merged.

PR #417, opened 2021-11-15
Feat/support webhook for pull request generator

shmurata:
As I wrote in https://github.com/argoproj-labs/applicationset/pull/366#issue-989590742, the pull request generator also supports webhooks for quick synchronization. I've added webhook support.

I also refactored some of the code because it was not good to simply add it. I also changed the annotations from the git generator-specific annotations to the common annotations, as it seemed to be a good idea to make the annotations common to all generators.
If you have any problems with this change, please comment and I will change it.


jgwest (approved):
Looks great, thanks @shmurata!

PR #423, opened 2021-11-26
fixes RepoHasPath error handling

benoitg31:
In the case that gitlab is unreachable , RepoHasPath would return false with NO error , thus filtering out the app from the "desired apps". This is not a wanted behaviour as the check could not be performed correctly.
The fixes checks the error from ListTree first, then checks the results.

### DIFF
--- a/pkg/services/scm_provider/gitlab.go
+++ b/pkg/services/scm_provider/gitlab.go
@@ -95,11 +95,11 @@ func (g *GitlabProvider) RepoHasPath(_ context.Context, repo *Repository, path s
 		Path: &path,
 		Ref:  &repo.Branch,
 	})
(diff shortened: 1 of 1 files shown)


yogeek:
Would be great to have this PR reviewed !
Indeed, this fix could solve the critical issue I explained in a [Slack discussion](https://cloud-native.slack.com/archives/C01U45M2SVB/p1636463565074700) with @jgwest.
I said "critical" because this issue cause some of our applications to be deleted randomly during our gitlab instance termination (and the linked resources are also deleted because we do not set the `preserveResourcesOnDeletion`).
Thank you 😀

jgwest (approved):
LGTM, thanks @benoitg31 and @yogeek!

PR #429, opened 2021-12-06
ApplicationSet Controller Panics when Cluster Secret is malformed

### LINKED ISSUE #398: ApplicationSet Controller Panics when Cluster Secret is malformed
Situation:  We found that the ApplicationSet controller v0.2.0 was in a crash loop back-off in my cluster.  The stack trace (will go looking for that in a moment) indicated an unexpected newline when unmarshaling the data.config, pointing to this line:

https://github.com/argoproj-labs/applicationset/blob/master/pkg/utils/clusterUtils.go#L107

The root cause was that the secret was created and the caBundle field was not base64 encoded.  Removing those secrets and restarting the ApplicationSet controller resolved the issue.

I'm wondering if a panic is an appropriate response, or just eliminate the affected cluster from the current compilation scope.

jgwest:
Fixes #398

### DIFF
--- a/pkg/utils/clusterUtils.go
+++ b/pkg/utils/clusterUtils.go
@@ -64,8 +64,13 @@ func ListClusters(ctx context.Context, clientset kubernetes.Interface, namespace
 	}
 	hasInClusterCredentials := false
 	for i, clusterSecret := range clusterSecrets {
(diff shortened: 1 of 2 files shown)


wtam2018 (approved):
LGTM
Thankls @jgwest !

PR #433, opened 2021-12-09
fix removed namespaces from resources on install

ekbduffy:
Some people may wish to install this chart not into "argocd" namespace and they will get errors then.

### DIFF
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -13606,7 +13606,6 @@ metadata:
     app.kubernetes.io/name: argocd-applicationset-controller
     app.kubernetes.io/part-of: argocd-applicationset
   name: argocd-applicationset-controller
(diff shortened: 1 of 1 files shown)


alexmt (changes requested):
I think install.yaml files are auto-generated from kustomize. Can you remove namespace from https://github.com/argoproj-labs/applicationset/blob/92623e6069851825d6b44be6421316b47626d9cb/manifests/base/kustomization.yaml#L9 

and 

https://github.com/argoproj-labs/applicationset/blob/92623e6069851825d6b44be6421316b47626d9cb/manifests/namespace-install/kustomization.yaml#L4

jgwest:
Fixed by https://github.com/argoproj/applicationset/pull/474

Issue #306, opened 2021-07-22
Sanitize application names


alvarogonzalez-packlink:
**Summary**: automatically generated application names (with a template variable, like `{{path.basename}}`, should be sanitized to avoid problematic characters.

**Background**: We have almost a hundred microservices with all its kustomize templates in a repository, so we've created an ApplicationSet using the git generator with this name template:
```yaml
[...]
  template:
    metadata:
      name: '{{path.basename}}'
      namespace: argocd
[...]
```
Everything seems to work properly until we notice the number of apps deployed by the application set doesn't match, and after some debugging we find the applications not deployed are the ones locates in directories with an underscore on its name.

Argo ingested the ApplicationSet seemingly ok, and there weren't any errors on Argo CD UI, but finally we found this in `argocd-applicationset-controller` logs:

```
2021-07-22T12:08:42.461Z        ERROR   controller-runtime.manager.controller.applicationset    Reconciler error    {"reconciler group": "argoproj.io", "reconciler kind": "ApplicationSet", "name": "production-apps-set", "namespace": "argocd", "error": "Application.argoproj.io \"office_back\" is invalid: metadata.name: Invalid value: \"office_back\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"}
github.com/go-logr/zapr.(*zapLogger).Error
        /go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/internal/controller/controller.go:267
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/internal/controller/controller.go:235
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1.1
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/internal/controller/controller.go:198
k8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext.func1
[...]

PR #436, opened 2021-12-14
fix: sanitize basename param by replacing unsupported character

missingcharacter:
Sanitize the cluster name in accordance with the below rules:

1. contain no more than 253 characters
2. contain only lowercase alphanumeric characters, '-' or '.'
3. start and end with an alphanumeric character

Closes https://github.com/argoproj-labs/applicationset/issues/306


thwiv (approved):

crenshaw-dev (changes requested):
How would you feel about adding a new variable called `path.basenameNormalized` to mirror the changes here? https://github.com/argoproj-labs/applicationset/pull/390

If we go the direction of adding a new variable, I think it should also be documented (along with the cluster generator variable if you have some extra time 😄 ).

@jgwest just pinging with this thought: if we were to introduce some more sophisticated templating tool (e.g. expr), this would be a use case.

docs/Generators-Cluster.md
```diff
@@ -2,11 +2,12 @@
...
-For each cluster registered with Argo CD, the Cluster generator produces parameters based on the list of items found within the cluster secret. 
+For each cluster registered with Argo CD, the Cluster generator produces parameters based on the list of items found within the cluster secret.
 
 It automatically provides the following parameter values to the Application template for each cluster:
 
 - `name`
+- `nameNormalized` *(In case the name of your cluster contains unsupported characters, this parameter will replace them with `-`)*
 - `server`
 - `metadata.labels.<key>` *(for each label in the Secret)*
 - `metadata.annotations.<key>` *(for each annotation in the Secret)*
```
thecubed, line 13: It might make sense to move the admonition about the unsupported characters to a doc themed callout after the parameter list:

```suggestion
- `name`
- `nameNormalized`
- `server`
- `metadata.labels.<key>` *(for each label in the Secret)*
- `metadata.annotations.<key>` *(for each annotation in the Secret)*

!!! note
    Use the `nameNormalized` parameter if your cluster name contains characters (such as underscores) that are not valid for Kubernetes resource names. This prevents rendering invalid Kubernetes resources with names like `my_cluster-app1`, and instead would convert them to `my-cluster-app1`.

```

docs/Generators-Git.md
```diff
@@ -55,6 +55,7 @@ The generator parameters are:
 
 - `{{path}}`: The directory paths within the Git repository that match the `path` wildcard.
 - `{{path.basename}}`: For any directory path within the Git repository that matches the `path` wildcard, the right-most path name is extracted (e.g. `/directory/directory2` would produce `directory2`).
+- `{{path.basenamenameNormalized}}`: Same as `path.basename` BUT unsupported characters will be replaced with `-` (e.g. `/directory/directory_2` would produce `directory-2`).
```
thecubed, line 58:
```suggestion
- `{{path.basenamenameNormalized}}`: This field is the same as `path.basename` with unsupported characters replaced with `-` (e.g. a `path` of `/directory/directory_2`, and `path.basename` of `directory_2` would produce `directory-2` here).
```

docs/Generators-Git.md
```diff
@@ -225,6 +226,7 @@ In addition to the flattened key/value pairs from the configuration file, the fo
 
 - `{{path}}`: The path to the folder containing matching configuration file within the Git repository. Example: `/clusters/clusterA`, if the config file was `/clusters/clusterA/config.json`
 - `{{path.basename}}`: Basename of the path to the folder containing the configuration file (e.g. `clusterA`, with the above example.)
+- `{{path.basenamenameNormalized}}`: Same as `path.basename` BUT unsupported characters will be replaced with `-`
```
thecubed, line 229:
```suggestion
- `{{path.basenamenameNormalized}}`: This field is the same as `path.basename` with unsupported characters replaced with `-` (e.g. a `path` of `/directory/directory_2`, and `path.basename` of `directory_2` would produce `directory-2` here).
```

Same as above

thecubed (approved):
Hi @missingcharacter !
Per our discussion earlier, here's some suggestions for the documentation.

This does seem to me like it might be useful for this project to switch to using a template generator that allows filters so one could write something like `{{ basename | replace(".","_") }}` akin to jinja2 or Go templates.

ApplicationSets appears to use https://github.com/valyala/fasttemplate (https://github.com/argoproj-labs/applicationset/blob/master/pkg/utils/util.go#L16) which doesn't support functions or filters, though.

That might be a task for someone else to take on at a later point though.

crenshaw-dev (approved):

missingcharacter:
Thanks @crenshaw-dev

crenshaw-dev:
@jgwest can you take a look?

missingcharacter:
@jgwest am I missing something for workflows to be approved?

wtam2018:
please resolve conflicts @missingcharacter

missingcharacter:
@wtam2018 I did, could you please approve the workflows?

Also, I ran my changes on my work and they seem good https://github.com/missingcharacter/applicationset/runs/4668321217?check_suite_focus=true

wtam2018 (approved):
LGTM
Thanks @missingcharacter

missingcharacter:
Thanks @wtam2018 🙇🏽‍♂️

PR #439, opened 2021-12-17
fix: log all validation errors

crenshaw-dev:
Logs are cheap. Logging all validation errors will make it easier to resolve them all.

Still only putting one error on the status to avoid using too much space.

### DIFF
--- a/pkg/controllers/applicationset_controller.go
+++ b/pkg/controllers/applicationset_controller.go
@@ -143,12 +143,12 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		var message string
 		for _, v := range validateErrors {
 			message = v.Error()
-			break
+			log.Errorf("validation error found during application validation: %s", message)
 		}
 		if len(validateErrors) > 1 {
+			// Only the last message gets added to the appset status, to keep the size reasonable.
 			message = fmt.Sprintf("%s (and %d more)", message, len(validateErrors)-1)
 		}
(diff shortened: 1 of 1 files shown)


pkg/controllers/applicationset_controller.go
```diff
@@ -143,12 +143,12 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		var message string
 		for _, v := range validateErrors {
 			message = v.Error()
-			break
+			log.Errorf("validation error found during application validation: %s", message)
 		}
 		if len(validateErrors) > 1 {
+			// Only the last message gets added to the appset status, to keep the size reasonable.
```
wtam2018, line 149: Should we keep the old behavior which adds the first error to appset status?

crenshaw-dev: @wtam2018 good question. It's pulling from an [unordered map](https://go.dev/blog/maps#iteration-order), so getting the last message is effectively the same as getting the first message.

crenshaw-dev: Maybe I should change the comment to "last encountered message" to clarify?

wtam2018: Thanks for the clarification. The comment is fine. It is an unordered map so we are making any claim as to which one of the errors to return.

wtam2018 (approved):
LGTM
thanks @crenshaw-dev

PR #444, opened 2021-12-22
chore: set applicationset-controller containerPort name

aabouzaid:
Hi,

A small change removes the redundancy of the port number by setting the applicationset-controller containerPort name.

It's better for readability as well as it minimizes the places to touch if the port is changed.

Thanks.


wtam2018 (approved):
LGTM
Thanks @aabouzaid

PR #451, opened 2022-01-06
fix: Set controller logger if we don't use JSON format

### LINKED ISSUE #445: Application set v0.3.0 log regression
When running:
```
docker run --rm -it quay.io/argoproj/argocd-applicationset:v0.3.0 applicationset-controller
```
It errors out, producing 0 logs
Compared to 
```
docker run --rm -it quay.io/argoproj/argocd-applicationset:v0.2.0 applicationset-controller
```

This makes debugging CI failures, such as https://github.com/argoproj/argo-helm/pull/1070 tricky.

mkilchhofer:
Resolves #445

### DIFF
--- a/main.go
+++ b/main.go
@@ -97,8 +97,8 @@ func main() {
 
 	json := strings.ToLower(logFormat) == JsonFormat
 
+	ctrl.SetLogger(zap.New(zap.UseDevMode(!json)))
 	if json {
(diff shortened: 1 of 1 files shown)


wtam2018 (approved):
LGTM
Thanks @mkilchhofer

PR #454, opened 2022-01-11
feat Gitlab list all repos

r0bobo:
This PR adds the feature for the Gitlab SCM generator to list all projects visible to the current user.
I was not sure if you would want to modify the CRD to make this more explicit so now it just lists all projects if `spec.generators.scmProvider.group` is an empty string and `spec.generators.scmProvider.includeSubgroups` is just ignored in this case.


r0bobo:
Are the failed Github unit tests due to a intermittent issue with Github? I couldn't get these to fail locally.

wtam2018:
You might want to wait for https://github.com/argoproj/applicationset/pull/455` to merge and sync.

rishabh625:
@r0bobo :  Can u please move this PR into argocd as applicationset is [moved](https://github.com/argoproj/argo-cd/pull/8864). If u find it laborious, i can do it on your behalf with your consent

PR #455, opened 2022-01-11
chore: update references from argoproj-labs to argoproj

ishitasequeira:
As part of moving application set from `argoproj-labs` to `argoproj`, we updated references from `argoproj-labs/applicationset` to `argoproj/applicationset` in the code base.

Will update CI and docs in a separate PR.

Signed off by [Ishita Sequeira](@ishitasequeira )


ishitasequeira:
@jgwest PTAL.

jgwest (changes requested):
Looks great, found a few more spots to update:
- `/docs` folder, seems like there are a few doc pages that are still using the old applicationset repo url
- `README.md` (a couple of references to old repo)
- (I noticed CHANGELOG.md had some references, but I think we can leave these as they are)

jgwest (approved):
Perfect, thanks again @ishitasequeira!

PR #469, opened 2022-01-20
Add pull request support to SCM generator

fardin01:
This commit adds pull request support to SCM generator so the generator
can create ArgoCD apps for PRs as well.

Fixes https://github.com/argoproj/applicationset/issues/466


fardin01:
Waiting for https://github.com/argoproj/applicationset/pull/472 to be merged.

api/v1alpha1/applicationset_types.go
```diff
@@ -342,6 +346,10 @@ type SCMProviderGeneratorFilter struct {
 	LabelMatch *string `json:"labelMatch,omitempty"`
 	// A regex which must match the branch name.
 	BranchMatch *string `json:"branchMatch,omitempty"`
+	// A regex which must match the pull request tile.
+	PullRequestBranchMatch *string `json:"pullRequestTitleMatch,omitempty"`
```
crenshaw-dev, line 350: Should there be two filters, one to match source branches, and the other to match target branches?

pkg/services/scm_provider/gitlab.go
```diff
@@ -60,6 +61,31 @@ func (g *GitlabProvider) GetBranches(ctx context.Context, repo *Repository) ([]*
...
+	}
+
+	for _, pullRequest := range pullRequests {
+		repos = append(repos, &Repository{
+			Organization: repo.Organization,
+			Repository:   repo.Repository,
+			URL:          repo.URL,
+			Branch:       pullRequest.Title,
+			SHA:          pullRequest.SHA,
+			Labels:       pullRequest.Labels,
```
crenshaw-dev, line 82: What if instead of setting the top-level labels field to be the PR labels, we added a `PullRequest` field with fields such as `SourceBranch`, `TargetBranch`, and `Labels`? That would leave the top-level labels field to have its usual meaning (repo labels).

The `PullRequest` field should also have an `ID` field to match what's provided by the pull request generator.

pkg/services/scm_provider/utils.go
```diff
@@ -114,10 +155,18 @@ func ListRepos(ctx context.Context, provider SCMProviderService, filters []argop
 		}
 	}
 
-	repos, err = getBranches(ctx, provider, filteredRepos, compiledFilters)
+	pullRequestRepos, err := getPullRequests(ctx, provider, filteredRepos, compiledFilters)
 	if err != nil {
 		return nil, err
 	}
+
+	branchRepos, err := getBranches(ctx, provider, filteredRepos, compiledFilters)
```
crenshaw-dev, line 163: I'm growing progressively more uncomfortable with the "repos" struct being used for non-repo stuff like pull requests and branches. But that's probably a refactor for another day. :-)

crenshaw-dev (changes requested):

pkg/services/scm_provider/github.go
```diff
@@ -104,7 +134,7 @@ func (g *GithubProvider) ListRepos(ctx context.Context, cloneProtocol string) ([
 
 func (g *GithubProvider) RepoHasPath(ctx context.Context, repo *Repository, path string) (bool, error) {
 	_, _, resp, err := g.client.Repositories.GetContents(ctx, repo.Organization, repo.Repository, path, &github.RepositoryContentGetOptions{
-		Ref: repo.Branch,
+		Ref: repo.SHA,
```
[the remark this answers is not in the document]
fardin01: So `repo.Branch` would only work when dealing with branches in a repository. On the other hand, using `repo.SHA` would work for both branches and pull requests (created using a fork in my case).

pkg/services/scm_provider/utils.go
fardin01: I completely agree. It did not feel 100% right doing this. How about renaming the `Repository` struct to `argoapp`, since that's basically what it ends up being?

pkg/services/scm_provider/gitlab.go
```diff
@@ -60,6 +61,31 @@ func (g *GitlabProvider) GetBranches(ctx context.Context, repo *Repository) ([]*
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, pullRequest := range pullRequests {
+		repos = append(repos, &Repository{
+			Organization: repo.Organization,
+			Repository:   repo.Repository,
+			URL:          repo.URL,
+			Branch:       pullRequest.Title,
```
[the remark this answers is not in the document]
fardin01: This is my mistake. I will change it to `pullRequest.SourceBranch` which is the name of the source branch. This would be consistent with what the Github provider does as well.

api/v1alpha1/applicationset_types.go
fardin01: That's an interesting point! I can't recall ever seeing a source branch being merged into a non-default target branch (main, master etc.). In addition, usually, all source branches get merged into one single target branch (otherwise you got a very messy repo 😄 ) which makes filtering the target branch pointless. So IMHO this change would only support a corner case. What do you think?

pkg/services/scm_provider/utils.go
```diff
@@ -52,6 +71,23 @@ func matchFilter(ctx context.Context, provider SCMProviderService, repo *Reposit
 		return false, nil
 	}
 
+	if filter.PullRequestBranchMatch != nil && !filter.PullRequestBranchMatch.MatchString(repo.Branch) {
```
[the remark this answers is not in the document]
fardin01: Good point. I'll change it to `if filter.BranchMatch != nil || filter.PullRequestBranchMatch != nil && ...`

pkg/services/scm_provider/gitlab.go
fardin01: This is a very good suggestion, I'll implement it. Also, this way we get to keep the dedicated `if` statement for `filter.PullRequestLabelMatch` in `matchFilter` function.

pkg/services/scm_provider/utils.go
```diff
@@ -52,6 +71,23 @@ func matchFilter(ctx context.Context, provider SCMProviderService, repo *Reposit
...
+		return false, nil
+	}
+
+	if filter.PullRequestLabelMatch != nil {
```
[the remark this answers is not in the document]
fardin01: With adding the `PullRequest` field to the `Repository` struct, we can keep this and range through `repo.PullRequest.Labels` instead of `repo.Labels`.
https://github.com/argoproj/applicationset/pull/469#discussion_r819411302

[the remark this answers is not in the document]
crenshaw-dev: Yep, I think that's the way to go.

Will need to consider how to treat a pull-request filter placed on an SCM generator where pull request fetching isn't enabled.

api/v1alpha1/applicationset_types.go
crenshaw-dev: Consider this use case: I want to spin up test environments for each PR against `main`. Developer A opens a PR from `feature/whatever` to `main`. Developer B has suggestions but instead of writing them all out, just opens a PR from `feature/whatever-review` to `feature/whatever`. 

I think this is common enough to justify giving the AppSet the designer to avoid creating an App for that second PR.

crenshaw-dev: But if it ends up being a lot of work, we could certainly leave it for a future enhancement.

pkg/services/scm_provider/utils.go
crenshaw-dev: Yep, I like that idea. Maybe something like `SCMGeneratorApp` to differentiate from the eventual actual App? idk, at your discretion. :-)

fardin01: I'll do this in a follow-up PR. 👍🏼

pkg/services/scm_provider/github_test.go
```diff
@@ -67,11 +67,17 @@ func TestGithubListRepos(t *testing.T) {
 			url:         "git@github.com:argoproj/applicationset.git",
 			branches:    []string{"master", "release-0.1.0"},
 		},
+		{
+			name:            "all pull requests",
+			allPullRequests: true,
+			url:             "git@github.com:argoproj/applicationset.git",
+			branches:        []string{"pr-1", "pr-2"},
```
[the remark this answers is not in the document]
fardin01: Okay, I misunderstood what this test actually does. So it gets real data from applicationset Github repo and tests against that. In that case, in order to test for pull requests, we'll have to open a "dummy" PR and keep it always open for this purpose. Does that sound alright?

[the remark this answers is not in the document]
fardin01: On second thought, we can do that with `allPullRequests` set to true and a filter to look for a specific PR. We can't realistically list all the PRs and check if their branch is what we expect them to be. We could set `allPullRequests` to true and check for the response code. How does that sound?

rishabh625:
@fardin01 :  Can u please move this PR into argocd as applicationset is [moved](https://github.com/argoproj/argo-cd/pull/8864). If u find it laborious, i can do it on your behalf with your consent

fardin01:
@rishabh625 Please feel free to move this PR over to argocd repo. I should have some time to get back at this pretty soon. Thanks :)

Issue #464, opened 2022-01-18
Github rate limit


fardin01:
The SCM Provider Generator is a great tool for discovering the repos that have something to deploy to Kubernetes. The only problem is, it makes a ton of API requests to Github and very quickly gets rate limited. My Github org has around 600 repos which I don't think is unusually high.

One place for optimization I believe is [this line](https://github.com/argoproj/applicationset/blob/master/pkg/services/scm_provider/github.go#L67). It seems that [list-branches](https://developer.github.com/v3/repos/#list-branches) API is called once for every repo in the org. If a `filter` is specified (which I believe is the case most of the time), then the filter should be taken into account here to avoid calling `list-branches` for repos that we are not interested in. I will look for more places that can be optimized.

Is there someone intersted in looking into this?

mmerrill3:
Would a topics filter help out here?  Let's say a repo has a topic of "agocd-target", and there was an option to filter on this topic?  It would be github.com specific though.

fardin01:
I think it would definitely help out. There is a `labelMatch` filter already, so we could just compare that with the topics of each repo (returned by `ListByOrg`) and if there is no match, we skip calling `list-branches` API.  Exact same thing can be done using `allBranches` attribute as well as `repositoryMatch` filter.

Like I said I have ~ 600 repos in my org, so it should take a maximum of [6 API calls](https://github.com/argoproj/applicationset/blob/377b7e5b816d4e7b792bf89484fa2d85c6f70609/pkg/services/scm_provider/github.go#L46-L51) to get most of the information we need.

enys:
I'm confirming this. Hitting also the problem.
It also seems the requeueAfterSeconds is not behaving as expected.

mmerrill3:
This logic also exists in the other scm_provider, gitlab.  The fix would be to add another function to the scm_provider interface to fill in the branches after the utils class filters the repos.  [Here](https://github.com/argoproj/applicationset/blob/c8fb714a35fc66805fdb184dd5e6455637d5a8b4/pkg/services/scm_provider/utils.go#L108), probably.  That's where the new function call would be to fill in the branches for all the scm_providers.

fardin01:
Sounds good @mmerrill3. Looking forward to seeing what you'll put together 🚀

fardin01:
@enys In the meantime, if you have Github Enterprise Cloud, you can create an OAuth token that has a 15k/hour rate limit, and in my case, it is JUST enough.

[...]

PR #472, opened 2022-01-20
(#464) Github rate limit easing

mmerrill3:
Signed-off-by: mmerrill3 <jjpaacks@gmail.com>

Fix for #464, lookup branches after applying repo filters.  Applies to github.com and gitlab implementations.

### DIFF
--- a/pkg/services/scm_provider/github.go
+++ b/pkg/services/scm_provider/github.go
@@ -42,6 +42,27 @@ func NewGithubProvider(ctx context.Context, organization string, token string, u
 	return &GithubProvider{client: client, organization: organization, allBranches: allBranches}, nil
 }
 
+func (g *GithubProvider) GetBranches(ctx context.Context, repo *Repository) ([]*Repository, error) {
+	repos := []*Repository{}
+	branches, err := g.listBranches(ctx, repo)
+	if err != nil {
+		return nil, fmt.Errorf("error listing branches for %s/%s: %v", repo.Organization, repo.Repository, err)
+	}
+
+	for _, branch := range branches {
+		repos = append(repos, &Repository{
+			Organization: repo.Organization,
+			Repository:   repo.Repository,
+			URL:          repo.URL,
+			Branch:       branch.GetName(),
+			SHA:          branch.GetCommit().GetSHA(),
+			Labels:       repo.Labels,
+			RepositoryId: repo.RepositoryId,
+		})
+	}
+	return repos, nil
+}
+
 func (g *GithubProvider) ListRepos(ctx context.Context, cloneProtocol string) ([]*Repository, error) {
 	opt := &github.RepositoryListByOrgOptions{
 		ListOptions: github.ListOptions{PerPage: 100},
@@ -63,22 +84,14 @@ func (g *GithubProvider) ListRepos(ctx context.Context, cloneProtocol string) ([
 			default:
 				return nil, fmt.Errorf("unknown clone protocol for GitHub %v", cloneProtocol)
 			}
-
-			branches, err := g.listBranches(ctx, githubRepo)
-			if err != nil {
-				return nil, fmt.Errorf("error listing branches for %s/%s: %v", githubRepo.Owner.GetLogin(), githubRepo.GetName(), err)
-			}
-
-			for _, branch := range branches {
-				repos = append(repos, &Repository{
-					Organization: githubRepo.Owner.GetLogin(),
-					Repository:   githubRepo.GetName(),
-					URL:          url,
-					Branch:       branch.GetName(),
-					SHA:          branch.GetCommit().GetSHA(),
-					Labels:       githubRepo.Topics,
-				})
-			}
+			repos = append(repos, &Repository{
+				Organization: githubRepo.Owner.GetLogin(),
+				Repository:   githubRepo.GetName(),
+				Branch:       githubRepo.GetDefaultBranch(),
+				URL:          url,
+				Labels:       githubRepo.Topics,
+				RepositoryId: githubRepo.ID,
+			})
 		}
 		if resp.NextPage == 0 {
 			break
@@ -102,10 +115,10 @@ func (g *GithubProvider) RepoHasPath(ctx context.Context, repo *Repository, path
 	return true, nil
 }
 
-func (g *GithubProvider) listBranches(ctx context.Context, repo *github.Repository) ([]github.Branch, error) {
+func (g *GithubProvider) listBranches(ctx context.Context, repo *Repository) ([]github.Branch, error) {
 	// If we don't specifically want to query for all branches, just use the default branch and call it a day.
 	if !g.allBranches {
-		defaultBranch, _, err := g.client.Repositories.GetBranch(ctx, repo.Owner.GetLogin(), repo.GetName(), repo.GetDefaultBranch())
+		defaultBranch, _, err := g.client.Repositories.GetBranch(ctx, repo.Organization, repo.Repository, repo.Branch)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


pkg/services/scm_provider/github.go
```diff
@@ -63,22 +84,14 @@ func (g *GithubProvider) ListRepos(ctx context.Context, cloneProtocol string) ([
...
-					URL:          url,
-					Branch:       branch.GetName(),
-					SHA:          branch.GetCommit().GetSHA(),
-					Labels:       githubRepo.Topics,
-				})
-			}
+			repos = append(repos, &Repository{
+				Organization: githubRepo.Owner.GetLogin(),
+				Repository:   githubRepo.GetName(),
+				Branch:       githubRepo.GetDefaultBranch(),
```
fardin01, line 90: `GetDefaultBranch()` would still call Github API for every repo even if there is a `repositoryMatch` or `labelMatch` filter.

At this point, we already do know the repo names and their labels/topics (from `ListByOrg`). So why not match the names and labels/topics with `repositoryMatch` or `labelMatch` and skip calling `GetDefaultBranch()` alltogether?

fardin01: As for the other filters (`pathsExist` and `branchMatch`), there is no way around it. Github API must be called for every repo to get that data 😞

mmerrill3: ok, thanks for that insight, I'll update that to use the default branch that is already present, as opposed to getting the default branch with another API call.  I didn't realize it was doing that.

mmerrill3: In regards to the two filters you mention, PathExists, and BranchMatch, those two filters will get applied now at two different phases.   The new function in utils.go will determine where a filter is for a branch, or for a repo.  PathExists is meant for a repo, and gets applied during ListRepos function.  The other filter, BranchMatch, gets applied only after we've applied other filters to the repos, whittling down the repos using either PathExists, or LabelMatch.  

Here is that function, which determines if a filter is for a branch, or repo:

```
func getApplicableFilters(filters []*Filter, branchFilter bool) []*Filter {
	applicableFilters := []*Filter{}
	for _, filter := range filters {
		if branchFilter && filter.BranchMatch != nil {
			applicableFilters = append(applicableFilters, filter)
		} else if !branchFilter && filter.BranchMatch == nil {
			applicableFilters = append(applicableFilters, filter)
		}
	}
	return applicableFilters
}
```

fardin01: Thank you very much for the explanation, it's clear now. Solid work 👍🏼

pkg/services/scm_provider/github.go
```diff
@@ -102,10 +115,10 @@ func (g *GithubProvider) RepoHasPath(ctx context.Context, repo *Repository, path
 	return true, nil
 }
 
-func (g *GithubProvider) listBranches(ctx context.Context, repo *github.Repository) ([]github.Branch, error) {
+func (g *GithubProvider) listBranches(ctx context.Context, repo *Repository) ([]github.Branch, error) {
 	// If we don't specifically want to query for all branches, just use the default branch and call it a day.
 	if !g.allBranches {
-		defaultBranch, _, err := g.client.Repositories.GetBranch(ctx, repo.Owner.GetLogin(), repo.GetName(), repo.GetDefaultBranch())
+		defaultBranch, _, err := g.client.Repositories.GetBranch(ctx, repo.Organization, repo.Repository, repo.Branch)
```
coderanger, line 121: Not all repos are owned by an organization, not sure why you changed these.

mmerrill3: Hi @coderanger, this function now handles a "scm_provider.Repository".  This is because the listing of branches now occurs after ListRepos, and after the repo level filters have been applied.  At this point inside the listBranches function, the githubRepository was found in ListRepos already, and that repo is transformed into an "scm_provider.Repository" as a return of that function call (ListRepos).  Since branches are obtained after ListRepos now, we are dealing with an "scm_provider.Repository" instead of a github.Repository type.

coderanger: D'oh, hoisted by my own petard then :) (I thought about not naming it Organization in the internal struct originally but figured it would cause more confusion).

fardin01:
@mmerrill3 Is this ready for review?

mmerrill3:
@fardin01 yes, ready to start review process.

fardin01:
@jgwest @wtam2018 Could you please review this very cool and needed enhancement?

pkg/services/scm_provider/github.go
crenshaw-dev: I think applying `PathsExist` at the repo-level filter might be a mistake. Even if the default branch doesn't have a certain path, a non-default branch might have that path.

crenshaw-dev (commented):
Actually, besides the one comment LGTM. I considered suggesting some refactoring, but I think it'd be more involved than this PR should take on. :-)

pkg/services/scm_provider/utils.go
```diff
@@ -107,3 +145,15 @@ func ListRepos(ctx context.Context, provider SCMProviderService, filters []argop
...
+	applicableFilters := []*Filter{}
+	for _, filter := range filters {
+		if branchFilter && filter.BranchMatch != nil {
+			applicableFilters = append(applicableFilters, filter)
+		} else if !branchFilter && filter.BranchMatch == nil {
+			applicableFilters = append(applicableFilters, filter)
+		}
+	}
+	return applicableFilters
+}
```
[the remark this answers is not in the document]
mmerrill3: 👍 I'll update this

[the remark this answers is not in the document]
mmerrill3: @crenshaw-dev, do we want to build into the Filter struct whether a filter is meant for a repo level or branch level?  I think that would make the code easier to read.

pkg/services/scm_provider/github.go
mmerrill3: I moved this to a branch level filter

crenshaw-dev:
@mmerrill3 taking another look at this in a few hours. I think my only suggestion is going to be changing getApplicableFilters to return `([]*BranchFilter, []*RepoFilter)` since I think that'll read a bit more nicely. But I might just do a follow-up PR to see if I think the extra lines of code are worth it. :-)

crenshaw-dev:
In the meantime, @wtam2018 or @jgwest can you take a look?

mmerrill3:
> @mmerrill3 taking another look at this in a few hours. I think my only suggestion is going to be changing getApplicableFilters to return `([]*BranchFilter, []*RepoFilter)` since I think that'll read a bit more nicely. But I might just do a follow-up PR to see if I think the extra lines of code are worth it. :-)

I was able to push a new update with some new enum types, to see whether a filter is branch or repo type, and to leave room for other types in the future.  It simplified the getApplicableFilters function.

mmerrill3:
@crenshaw-dev I was thinking more on this, and I decided to return a map from getApplicableFilters, and I added a unit test to validate the behavior.  I'll let you know when its ready for review again.

pkg/services/scm_provider/utils.go
```diff
@@ -18,21 +18,25 @@ func compileFilters(filters []argoprojiov1alpha1.SCMProviderGeneratorFilter) ([]
...
 		if filter.PathsExist != nil {
 			outFilter.PathsExist = filter.PathsExist
+			outFilter.FilterType = FilterTypeBranch
 		}
 		if filter.BranchMatch != nil {
 			outFilter.BranchMatch, err = regexp.Compile(*filter.BranchMatch)
 			if err != nil {
 				return nil, fmt.Errorf("error compiling BranchMatch regexp %q: %v", *filter.LabelMatch, err)
 			}
+			outFilter.FilterType = FilterTypeBranch
```
crenshaw-dev, line 39: I think a problem with this implementation is that if there's a filter including both a repo filter and a branch filter, the whole filter will get marked as a branch filter.

crenshaw-dev (changes requested):

pkg/services/scm_provider/utils.go
mmerrill3: thanks, will check later today.

wtam2018 (approved):
LGTM

PR #477, opened 2022-01-25
docs: be explicit about which namespace appsets should be in (#476)

### LINKED ISSUE #476: docs should be more explicit about necessity of appset namespace
Users aren't realizing that appset resources need to be in the same ns as the controller (like [here](https://cloud-native.slack.com/archives/C01U45M2SVB/p1642616600045200) and [here](https://github.com/argoproj/applicationset/issues/456)).

Make sure all examples have a note about using the correct namespace.

crenshaw-dev:
Fixes #476 

I think hard-coding the `argocd` namespace will clarify more than confuse. People who are using non-standard namespaces will be more likely to know that they need to adjust these manifests.


ishitasequeira (approved):
The change LGTM!!.

wtam2018 (approved):

PR #504, opened 2022-02-15
docs: Updating cascading strategy

jtyr:
As per the [documentation](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#non-cascading-delete), if a resource should be deleted without deleting depending resources, the `orphan` cascading strategy should be used.


crenshaw-dev (approved):

ishitasequeira (approved):
LGTM

wtam2018 (approved):

PR #506, opened 2022-02-15
feat: add support for Bitbucket Server (SCM + PR)

### LINKED ISSUE #505: Add support for Bitbucket Server
There's already an issue for Bitbucket Cloud (https://github.com/argoproj/applicationset/issues/459), but not for Bitbucket Server.
The REST API is also different, I think it would be simpler to threat them as separate SCMs

mlosicki:
#### Implementation details
Uses Basic auth instead of access tokens like the other SCMs. Note that [Personal Access Tokens](https://confluence.atlassian.com/bitbucketserver0717/personal-access-tokens-1087535496.html#Personalaccesstokens-usingpersonalaccesstokens) do work with this.

PRs do not support labels. Added a `branchMatch` field for PRs which allows you to filter based on the source branch name (e.g. only consider PRs that include `-argocd`)

Repos do indeed support labels, but the `go-bitbucket-v1` library does not. It would have to be added upstream first.

Speaking of `go-bitbucket-v1`, version was increased as it has some bugfixes that were merged and are related to this PR (e.g. missing pagination options)

Closes #505


docs/Generators-Pull-Request.md
```diff
@@ -53,6 +53,43 @@ spec:
...
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: myapps
+spec:
+  generators:
+  - pullRequest:
+      bitbucketServer:
```
mlosicki, line 68: Need to add missing doc for `branchMatch`

crenshaw-dev (changes requested):
In the tests, can you replace `assert.Nil(t, err)` with `assert.NoError(t, err)` and `context.TODO()` with `context.Background()`? They're just semantic differences, but should read just a tiny bit nicer.

pkg/services/pull_request/bitbucket_server.go
```diff
@@ -0,0 +1,97 @@
...
+
+	pullRequests := []*PullRequest{}
+	for {
+		response, err := b.client.DefaultApi.GetPullRequestsPage(b.projectKey, b.repositorySlug, paged)
+		if err != nil {
+			return nil, fmt.Errorf("error listing pull requests for %s/%s: %v", b.projectKey, b.repositorySlug, err)
+		}
+		pulls, err := bitbucketv1.GetPullRequestsResponse(response)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing pull request response %s: %v", response.Values, err)
```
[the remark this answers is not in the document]
mlosicki: What kind of secrets are you thinking of ? Here's an example of a response from the [docs (`/rest/api/1.0/projects/{projectKey}/repos/{repositorySlug}/pull-requests`)](https://docs.atlassian.com/bitbucket-server/rest/7.17.5/bitbucket-rest.html#idp300):
<details>

```
{
    "size": 1,
    "limit": 25,
    "isLastPage": true,
    "values": [
        {
            "id": 101,
            "version": 1,
            "title": "Talking Nerdy",
            "description": "It’s a kludge, but put the tuple from the database in the cache.",
            "state": "OPEN",
            "open": true,
            "closed": false,
            "createdDate": 1359075920,
            "updatedDate": 1359085920,
            "fromRef": {
                "id": "refs/heads/feature-ABC-123",
                "repository": {
                    "slug": "my-repo",
                    "name": null,
                    "project": {
                        "key": "PRJ"
                    }
                }
            },
            "toRef": {
                "id": "refs/heads/master",
                "repository": {
                    "slug": "my-repo",
                    "name": null,
                    "project": {
                        "key": "PRJ"
                    }
                }
            },
            "locked": false,
            "author": {
                "user": {
                    "name": "tom",
                    "emailAddress": "tom@example.com",
                    "id": 115026,
                    "displayName": "Tom",
                    "active": true,
                    "slug": "tom",
                    "type": "NORMAL"
                },
                "role": "AUTHOR",
                "approved": true,
                "status": "APPROVED"
            },
            "reviewers": [
                {
                    "user": {
                        "name": "jcitizen",
                        "emailAddress": "jane@example.com",
                        "id": 101,
                        "displayName": "Jane Citizen",
                        "active": true,
                        "slug": "jcitizen",
                        "type": "NORMAL"
                    },
                    "lastReviewedCommit": "7549846524f8aed2bd1c0249993ae1bf9d3c9998",
                    "role": "REVIEWER",
                    "approved": true,
                    "status": "APPROVED"
                }
            ],
            "participants": [
                {
                    "user": {
                        "name": "harry",
                        "emailAddress": "harry@example.com",
                        "id": 99049120,
                        "displayName": "Harry",
                        "active": true,
                        "slug": "harry",
                        "type": "NORMAL"
                    },
                    "role": "PARTICIPANT",
                    "approved": true,
                    "status": "APPROVED"
                },
                {
                    "user": {
                        "name": "dick",
                        "emailAddress": "dick@example.com",
                        "id": 3083181,
                        "displayName": "Dick",
                        "active": true,
                        "slug": "dick",
                        "type": "NORMAL"
                    },
                    "role": "PARTICIPANT",
                    "approved": false,
                    "status": "UNAPPROVED"
                }
            ],
            "links": {
                "self": [
                    {
                        "href": "http://link/to/pullrequest"
                    }
                ]
            }
        }
    ],
    "start": 0
}
```

</details>
If you prefer to avoid logging the email, I can log the same msg as 4 lines above i.e. the query params.

pkg/services/pull_request/bitbucket_server.go
```diff
@@ -0,0 +1,97 @@
...
+	return newBitbucketService(ctx, bitbucketConfig, projectKey, repositorySlug, branchMatch)
+}
+
+func NewBitbucketServiceNoAuth(ctx context.Context, url, projectKey, repositorySlug string, branchMatch *string) (PullRequestService, error) {
+	return newBitbucketService(ctx, bitbucketv1.NewConfiguration(url), projectKey, repositorySlug, branchMatch)
+}
+
+func newBitbucketService(ctx context.Context, bitbucketConfig *bitbucketv1.Configuration, projectKey, repositorySlug string, branchMatch *string) (PullRequestService, error) {
+	if !strings.HasSuffix(bitbucketConfig.BasePath, "/rest") {
+		bitbucketConfig.BasePath = bitbucketConfig.BasePath + "/rest"
```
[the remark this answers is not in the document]
mlosicki: The library is more-or-less auto-generated and follows this example for all methods: `a.client.cfg.BasePath + "/api/1.0/projects/{projectKey}/repos"`
I believe they used the WADL provided by Bitbucket which defines the base path like this `base="http://example.com/rest/"`. This basically means that `/rest` is mandatory. Initially, I also wrote it explicitly (it's even in the example provided), but decided to handle it in code because it basically always has to be there. Ideally, it would've been part of the library.

I don't mind removing this and adding some extra docs for this, but do note that it always needs to be there and "regular" scm users wouldn't know this detail without the example in the docs.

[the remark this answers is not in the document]
mlosicki: Sorry for the response delay, refactored it a bit and added support for `/rest/`

pkg/services/pull_request/bitbucket_server.go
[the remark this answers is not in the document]
mlosicki: Done, now the payload is logged via `log.Errorf` here and in all the other places. Not sure if you had another log level in mind (debug?)

api/v1alpha1/applicationset_types.go
```diff
@@ -367,6 +381,28 @@ type PullRequestGeneratorGithub struct {
...
+// PullRequestGenerator defines connection info specific to BitbucketServer.
+type PullRequestGeneratorBitbucketServer struct {
+	// Project to scan. Required.
+	Project string `json:"project"`
+	// Repo name to scan. Required.
+	Repo string `json:"repo"`
+	// The Bitbucket REST API URL to talk to e.g. https://bitbucket.org/rest Required.
+	API string `json:"api"`
+	// A regex which must match the branch name.
+	BranchMatch *string `json:"branchMatch,omitempty"`
```
crenshaw-dev, line 393: Instead of adding a `BranchMatch` filter here, could you add a `Filters []PullRequestGeneratorFilter` field to the `PullRequestGenerator` struct, to parallel the way filtering is done in `SCMProviderGenerator`? That will make things a bit more complex, but it'll make things more intuitive for the user and make sure GitHub users get filtering as well.

If you'd rather just drop this field and add filtering support on another PR, that might help get basic support in more quickly.

docs/Generators-SCM-Provider.md
```diff
@@ -94,6 +94,47 @@ For label filtering, the repository tags are used.
...
+        allBranches: true
+        # Credentials for Basic authentication. Required for private repositories.
+        basicAuth:
+          # The username to authenticate with
+          username: myuser
+          # Reference to a Secret containing the password or personal access token.
+          passwordRef:
+            secretName: mypassword
+            key: password
+        # Support for labels is TODO. Bitbucket server labels are not supported for PRs, but they are for repos
```
crenshaw-dev, line 123:
```suggestion
        # Support for filtering by labels is TODO. Bitbucket server labels are not supported for PRs, but they are for repos
```

crenshaw-dev (changes requested):

api/v1alpha1/applicationset_types.go
mlosicki: If I just follow how it's done for `SCMProviderGenerator` and write it as :
```go
type PullRequestGeneratorFilter struct {
	LabelMatch *string `json:"labelMatch,omitempty"`
	BranchMatch *string `json:"branchMatch,omitempty"`
}
```
adding a label filter for Bitbucket would be valid yaml, but it isn't supported. I should make a `PullRequestGeneratorBitbucketServerFilter` struct with only the `BranchMatch` then ? I want to keep the branch regex in one way or another, there's 0 way do to filtering for Bitbucket PRs otherwise. I'm not planning on adding branch name filtering for Github PRs here though.

Btw, I didn't do it like that since the Github PR generator also has a `labels` array at the "top" level and not something like this, which would be in line with SCM provider:
```yaml
- pullRequest:
      github:
          filters:
          - labelMatch: preview
```
But I agree that it should be unified ideally.

crenshaw-dev: I'd recommend this:

```go
type PullRequestGeneratorFilter struct {
	BranchMatch *string `json:"branchMatch,omitempty"`
}
```

If someone adds github PR label filtering support later, I think the controller can throw a warning if someone adds a label filter to a bitbucketserver pull request generator. I think that missing out on apply-time validation is worth it for keeping the schema similar to the SCM generator.

pkg/services/scm_provider/bitbucket_server.go
```diff
@@ -0,0 +1,221 @@
...
+		response, err := b.client.DefaultApi.GetContent_0(repo.Organization, repo.Repository, path, opts)
+		if response != nil && response.StatusCode == 404 {
+			// File not found
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		// 200 ok
+		return true, nil
```
crenshaw-dev, line 130: Do we know that the library returns an error for every non-200 code?

pkg/services/scm_provider/bitbucket_server.go
```diff
@@ -0,0 +1,221 @@
...
+	response, err := b.client.DefaultApi.StreamFiles_42(repo.Organization, repo.Repository, path, opts)
+	if response != nil && response.StatusCode == 404 {
+		// The path requested does not exist at the supplied commit.
+		return false, nil
+	}
+	if response != nil && response.StatusCode == 400 {
+		// If the path is a file, the first call will return 400: The path requested is not a directory at the supplied commit.
+		// Simply retry with an API call that works with files and expect a 200 return code
+		opts["type_"] = true // Only request the type, we don't need the content
+		response, err := b.client.DefaultApi.GetContent_0(repo.Organization, repo.Repository, path, opts)
```
crenshaw-dev, line 121: I'm sure you thought of this, but why not just start with this call?

pkg/services/scm_provider/bitbucket_server.go
```diff
@@ -0,0 +1,221 @@
...
+		if err != nil {
+			return false, err
+		}
+		// 200 ok
+		return true, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
```
crenshaw-dev, line 135: Ditto, do we know that no error necessarily means yes-directory? (I'm thinking of redirects, for example, as maybe being something where `nil != nil` but still does not imply success.)

crenshaw-dev (commented):
Just clarifying questions at this point. Overall LGTM, will be happy to approve after clarifications!

pkg/services/scm_provider/bitbucket_server.go
mlosicki: It does, the library generated code is like this:
```go
	if localVarHTTPResponse.StatusCode >= 300 {
		bodyBytes, _ := ioutil.ReadAll(localVarHTTPResponse.Body)
		return NewAPIResponseWithError(localVarHTTPResponse, bodyBytes, reportError("Status: %v, Body: %s", localVarHTTPResponse.Status, bodyBytes))
	}
```
which then always returns an error (possibly a different one if parsing fails)

pkg/services/scm_provider/bitbucket_server.go
mlosicki: Can't think of a case where it would not work, added a test with a redirect

pkg/services/scm_provider/bitbucket_server.go
mlosicki: Good question 😄 I added this fix afterwards without seeing the forest for the trees. It does indeed work with directories too. Refactored to only make this call instead.

crenshaw-dev:
@mlosicki this morning got taken up by another review. But I should have a chance to do a last pass review no later than tomorrow morning.

crenshaw-dev (approved):
This looks great! Thanks for your patience on the review. 🙂

rishabh625:
@mlosicki :  Can u please move this PR into argocd as applicationset is [moved](https://github.com/argoproj/argo-cd/pull/8864). If u find it laborious, i can do it on your behalf with your consent

rishabh625:
This is merged we can close it

simcax:
@mlosicki - A question about this PR. I can see that Basic Auth is implemented. We are using ssh with privatekeys stored in secrets towards bitbucket server. Do you know if this would work, or would we still need to provide basic auth as well?

mlosicki:
> @mlosicki - A question about this PR. I can see that Basic Auth is implemented. We are using ssh with privatekeys stored in secrets towards bitbucket server. Do you know if this would work, or would we still need to provide basic auth as well?

@simcax No, it won't work. Bitbucket Server doesn't have an option to authenticate to the REST API via ssh keys. Those keys are only used for Git operations. So you do need Basic auth, preferably via personal access tokens.

simcax:
> > @mlosicki - A question about this PR. I can see that Basic Auth is implemented. We are using ssh with privatekeys stored in secrets towards bitbucket server. Do you know if this would work, or would we still need to provide basic auth as well?
> 
> @simcax No, it won't work. Bitbucket Server doesn't have an option to authenticate to the REST API via ssh keys. Those keys are only used for Git operations. So you do need Basic auth, preferably via personal access tokens.

Ah - of coures, that makes sense - although a bit of a nuisance having to create a service user of sorts. Thx!

simcax:
Has this PR been moved to applicationSet controller in ArgoCD, now that the applicationSet Controller is part of it?

simcax:
> Yep! [argoproj/argo-cd#9049](https://github.com/argoproj/argo-cd/pull/9049)

Cool!

PR #509, opened 2022-02-16
chore: remove dependency on unmaintained 'errors' package (#490)

### LINKED ISSUE #490: github.com/pkg/errors is no longer maintained
pkg/errors is no longer maintained. See pkg/errors#245

We can replaces all calls to errors.Wrap/f with fmt.Errorf and removes the unneeded dependency on
github.com/pkg/errors since we don't support go 1.12 or lower. See microsoft/go-winio#227 for an example.

crenshaw-dev:
Fixes #490

### DIFF
--- a/test/e2e/fixture/applicationsets/utils/fixture.go
+++ b/test/e2e/fixture/applicationsets/utils/fixture.go
@@ -111,9 +111,7 @@ func EnsureCleanState(t *testing.T) {
 		}
 		if list != nil && len(list.Items) > 0 {
 			// Fail
-			msg := fmt.Sprintf("Waiting for list of ApplicationSets to be size zero: %d", len(list.Items))
-			// Intentionally not making this an Errorf, so it can be printf-ed for debugging purposes.
-			return fmt.Errorf(msg)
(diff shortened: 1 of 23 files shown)


blkperl (approved):

ishitasequeira (approved):
LGTM

test/e2e/fixture/applicationsets/utils/fixture.go
```diff
@@ -114,7 +113,7 @@ func EnsureCleanState(t *testing.T) {
 			// Fail
 			msg := fmt.Sprintf("Waiting for list of ApplicationSets to be size zero: %d", len(list.Items))
 			// Intentionally not making this an Errorf, so it can be printf-ed for debugging purposes.
-			return errors.New(msg)
+			return fmt.Errorf(msg)
```
wtam2018, line 116: Should we clean this (and other ones in this file) up to combine them into a single fmt.Errorf() call?  If we debug with printf, we would've to edit it anyway.

crenshaw-dev: Agreed, will update!

wtam2018 (approved):
LGTM
thanks @crenshaw-dev !

PR #511, opened 2022-02-18
chore: simpler filter segmentation

crenshaw-dev:


crenshaw-dev:
@mmerrill3 can you take a look?

fardin01 (approved):
The fix worked as expected in my cluster.

oleg-glushak:
I tried this code for a combination of `repositoryMatch` and `branchMatch` filters and it seems it doesn't work as expected. Only a default branch gets discovered, but not others.

crenshaw-dev:
@oleg-glushak sounds like I need to write some more tests...

rishabh625:
@crenshaw-dev : Can u move this to argocd and close here?

PR #514, opened 2022-02-20
feat: add exclude files when using the git file generator (#468)

### LINKED ISSUE #468: Add the ability to exclude files when using the git file generator
When using the git directory generator it is possible to exclude specific paths like this

```  generators:
  - git:
      repoURL: https://github.com/argoproj-labs/applicationset.git
      revision: HEAD
      directories:
      - path: examples/git-generator-directory/excludes/cluster-addons/*
      - path: examples/git-generator-directory/excludes/cluster-addons/exclude-helm-guestbook
        exclude: true
```

When using the git file generator I would like to be able to add files using a wildcard and then exclude specific files like this

```
spec:
  generators:
  - git:
      repoURL: https://github.com/argoproj-labs/applicationset.git
      revision: HEAD
      files:
      - path: "examples/git-generator-files-discovery/cluster-config/**/config.json"
      - path  "examples/git-generator-files-discovery/cluster-config/staging/config.json"
        exclude: true
```

Of course if there is another way to do this that I am not aware of then I would be happy to use that. If not I am also happy to implement this change if it is agreed.

adamjohnson01:
This adds the ability to exclude specific paths when using the git file generator.

Closes #468


templarfelix (approved):

pkg/generators/git.go
```diff
@@ -174,29 +177,36 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
...
-					WithField("appPath", appPath).Error("error while matching appPath to requestedPath")
-				continue
+	for _, itemPath := range allPaths {
+		include := false
+		exclude := false
+		for _, requestedPath := range items {
+			var match bool
+			var err error
+			if pathType == "files" {
+				pathGlob := glob.MustCompile(requestedPath.Path)
```
crenshaw-dev, line 189: If it's important to use glob instead of `path.Match`, I think we should use a `glob` field instead of the `path` field, to avoid confusion.

If it's not necessary to use glob, I'd recommend we just use `path.Match`.

crenshaw-dev (changes requested):
Thanks for working on this! Just one comment for now. Also, could you update the git generator docs page to show how to use the new feature?

pkg/generators/git.go
adamjohnson01: @crenshaw-dev, we need to use glob if the path contains "**", I will add a conditional that checks if the path contains that and then uses glob but falls back to path.Match if it does not.

crenshaw-dev: That is a cool trick... Could you apply it to the directory generator as well? I usually don't like implicit toggles, but I think if we add a line to the documentation about it, it'll be fine.

adamjohnson01: Both the file and directory generator will use the same method for filtering so it will be applied to both.

adamjohnson01: I have pushed the changes and updated the examples and docs

crenshaw-dev (changes requested):
Looking good! My only remaining concern is that `a/**/b.json` might be a valid path in `path.Match`, and we might be changing the behavior for any users who have that kind of pattern. I don't have time to investigate at the moment. 

Are you strongly opposed to just introducing a new `glob` field for both files and directories? I feel like that's safer and easier for users to understand.

adamjohnson01:
@crenshaw-dev, I misunderstood what you meant about adding a new glob field. I think that would be more confusing. The only time glob will need to be used is when there is a "**" and exclude is true. The '**' works when generating the directory and file lists as under the covers is it just the git binary that is run. The path.Match function does not translate '**' the same way that the lsFIles function that generates the file and directory lists does so the current setup is producing unexpected behaviour for anyone using '**' in excludes. I think that adding the '**' check could potentially break some existing config but the config that it 'breaks' is not working as it should anyway. Changing it will make the exclude path matching consistent with the file and directory path generation which will make troubleshooting issues with excludes a lot easier.

crenshaw-dev:
@adamjohnson01 thanks for the explanation! I don't think I'll be able to give this the review it deserves until Monday. Thanks for your patience!

adamjohnson01:
@crenshaw-dev, those are valid points. How about using the [doublestar](https://github.com/bmatcuk/doublestar) library instead of glob. It implements support for double star `**` matches for path.Match. The [patterns](https://github.com/bmatcuk/doublestar/blob/master/match.go#L8) are the same other than `**` support.

crenshaw-dev:
I do like doublestar a lot more! Looks better-maintained. But I don't think it resolves the first three points.

I've been using these tests to compare behavior:

```go
func TestGlobVsMatch(t *testing.T) {
	cases := []struct {
		name string
		testPath string
		pattern string
		expected bool
	}{
		{
			name: "double-star does not match separator",
			testPath: "/a/b/c",
			pattern: "/a/**",
			expected: true,
		},
		{
			name: "{ is a valid part of a path",
			testPath: "/a/{filename}",
			pattern: "/a/{filename}",
			expected: true,
		},
	}

	for _, testCase := range cases {
		testCaseCopy := testCase

		t.Run(testCaseCopy.name, func(t *testing.T) {
			t.Parallel()
			pathMatches, err := path.Match(testCaseCopy.pattern, testCaseCopy.testPath)
			assert.NoError(t, err)
			assert.Equal(t, testCaseCopy.expected, pathMatches, "path.Match returned %v instead of %v for pattern %v and path %v", pathMatches, testCaseCopy.expected, testCaseCopy.pattern, testCaseCopy.testPath)
			doublestarMatches, err := doublestar.Match(testCaseCopy.pattern, testCaseCopy.testPath)
			assert.NoError(t, err)
			assert.Equal(t, testCaseCopy.expected, doublestarMatches, "doublestar.Match returned %v instead of %v for pattern %v and path %v", doublestarMatches, testCaseCopy.expected, testCaseCopy.pattern, testCaseCopy.testPath)
		})
	}
}
```

adamjohnson01:
@crenshaw-dev 

> I do like doublestar a lot more! Looks better-maintained. But I don't think it resolves the first three points.

I think that 1 is the only point that it does not solve but this is limited to excludes and not the path generation as that is handled by `git ls-files`. The other points are no longer valid as we will no longer be using glob, we will be using path.Match again.

crenshaw-dev:
> this is limited to excludes and not the path generation as that is handled by git ls-files

Can you clarify what you mean here? I'm sure I'm missing something obvious. But as far as I can tell, file listing is currently performed by cloning the repo and using `filepath.Walk` to generate a list of files. 

So the behavior of this generator would change after merging this PR:

```yaml
  - git:
      repoURL: https://github.com/argoproj/applicationset.git
      revision: HEAD
      files:
      - path: "examples/git-generator-files-discovery/cluster-config/**/config.json"
```

Instead of matching only files one directory away from `cluster-config`, this would seem to now match config.json files in further-nested directories.

crenshaw-dev (approved):
LGTM! Thanks for walking me through understanding the implementation. Took me a while to understand fully. 😄

rishabh625:
@adamjohnson01 :  Can u please move this PR into argocd as applicationset is [moved](https://github.com/argoproj/argo-cd/pull/8864). If u find it laborious, i can do it on your behalf with your consent

adamjohnson01:
@rishabh625, I will give it a go

crenshaw-dev:
@adamjohnson01 let us know if you'd rather us move over the PR on your behalf.  :-)

yebolenko:
any progress on this PR?
When it will be merged?

rishabh625:
@yebolenko it's already merged and will be available in next release https://github.com/argoproj/argo-cd/pull/9150

bitfactory-henno-schooljan:
> @yebolenko it's already merged and will be available in next release [argoproj/argo-cd#9150](https://github.com/argoproj/argo-cd/pull/9150)

AFAICS this merge is related to excluding entire repositories using the SCM generator, how would I use this to exclude only a single file using the Git file generator when generating all Applications from a single Git repository?

adamjohnson01:
I have not had time to work on this but will try and get it moved across.

bitfactory-henno-schooljan:
Has there been any progress on this? I would like to ignore certain directory patterns but keep the rest, which is impossible using only glob patterns for inclusion as it is now.

vinicius-deoliveira:
Hi folks, 

Do we have some update about this? I would like to use exclude files when using git file generator.

PR #524, opened 2022-03-02
Gitlab list archived option

r0bobo:
This adds the option to exclude archived repos in the Gitlab SCM generator.  I added the option as `excludeArchived` to not change the current default and to not having to muck about with changing the CRD default to true somehow.


rishabh625:
@r0bobo : can you please move this changes to https://github.com/argoproj/argo-cd

PR #526, opened 2022-03-02
feat: add linux/arm64 support

glyphtech-chrisa:
plus `go mod tidy` required fix.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -97,6 +97,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
+	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/prometheus/client_golang v1.11.0 // indirect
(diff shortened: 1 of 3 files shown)


glyphtech-chrisa:
Context: Deploying ArgoCD via Helm to arm64 hosts and fails pulling the image for this. I've tested the arm64 image some locally and appears to work in K8s 1.23.

glyphtech-chrisa:
Hi Team, please review/approve when you get a chance: @crenshaw-dev @rishabh625 @jgwest @ishitasequeira @mmerrill3.

go.mod
```diff
@@ -97,6 +97,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
+	github.com/pkg/errors v0.9.1 // indirect
```
rishabh625, line 100: This fix is already present in master branch, how come diff shows it as new addition

crenshaw-dev: I think merging master would remove it from the diff. But it shouldn't cause any problems as-is.

glyphtech-chrisa: I fixed this and put out the PR before that was merged into master I believe.

wtam2018 (approved):
LGTM
thanks @glyphtech-chrisa !

PR #527, opened 2022-03-03
fix: Adding indirect dependency for errors pkg

rishabh625:
There is indirect dependency for error pkg 
go mod tidy - added this dependecy adding to fix e2e test

Signed-off-by: rishabh625 <rishabhmishra625@gmail.com>


wtam2018 (approved):
LGTM
thanks @rishabh625

PR #529, opened 2022-03-04
chore: Added 60s sleep between start e2e and test e2e,increased test …

rishabh625:
**What this PR does:**

This PR increases the e2e test run timeout from 20 mins to 25 mins.

Adding a longer delay between start e2e and test e2e could solve the current "master cron" e2e test failure. The assumption is that the Github runner become slow sometimes and the delay is not longer enough to allow  e2e test application to start.

### DIFF
--- a/.github/workflows/cron-master.yaml
+++ b/.github/workflows/cron-master.yaml
@@ -64,8 +64,9 @@ jobs:
           "hack/ci-e2e-setup.sh"
 
       - name: Run E2E tests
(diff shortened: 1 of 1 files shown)


rishabh625:
> Please provide more explanation/description about this PR.

Adding delay between start e2e and rest e2e , assuming that might be github runner become slow sometimes.

wtam2018 (approved):
LGTM

PR #533, opened 2022-03-08
fix: downgrade Argo CD dependency to exclude argoproj/argocd#ca2e3041

alexmt:

The https://github.com/argoproj/argo-cd/pull/8516 introduced an incompatible change in Argo CD that changes how repo credentials are passed to git. In order to accommodate this change, we would have to add `argocd` binary into applicationset image. This is too intrusive change, especially since we agreed to merge applicationset and argocd code. 

So I propose to temporary use pre v2.3 release version of Argo CD that does not have https://github.com/argoproj/argo-cd/pull/8516 and prioritize merging Application + Argo CD during v2.4 Argo CD .

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/argoproj/applicationset
 go 1.17
 
 require (
-	github.com/argoproj/argo-cd/v2 v2.3.0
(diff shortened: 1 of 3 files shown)


crenshaw-dev (approved):
LGTM!

wtam2018 (approved):
LGTM

Issue #386, opened 2021-10-05
Gitlab Generator: PathExists does not work for files


dsch:
# Current behavior

The [filter example](https://argocd-applicationset.readthedocs.io/en/stable/Generators-SCM-Provider/#filters) from the doc does not work for Gitlab SCM generator.

Only directories are supported as values for `pathsExist` values.

Example:
`pathsExist: [kubernetes]` works
`pathsExist: [kubernetes/kustomization.yaml]` does not work

# Expected behavior

Paths to files should match the filter.

# Background

GitlabProvider [RepoHasPath()](https://github.com/argoproj-labs/applicationset/blob/master/pkg/services/scm_provider/gitlab.go#L89) uses [List repository tree](https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree) Gitlab API with option path. From the docs: _The path inside repository. Used to get content of subdirectories._

daydy16:
Doesnt work for me either with specific files. 
Also it looks like labelMatch isn't working as well.

PR #540, opened 2022-03-11
fix: Gitlab Generator PathExists for files

dsch:
This is a fix for #386.

Unfortunately the Gitlab repository (https://gitlab.com/test-argocd-proton/argocd) used for the tests does not contain enough files the test the pagination.

_Note:_ This is my first PR in golang. Feel free do propose changes to make it more the "Go-way".


pkg/services/scm_provider/gitlab.go
```diff
@@ -112,10 +113,34 @@ func (g *GitlabProvider) RepoHasPath(_ context.Context, repo *Repository, path s
...
-	return true, nil
+
+	// let's try if path is a file
+	dir := pathpkg.Dir(path)
+	options := gitlab.ListTreeOptions{
+		Path: &dir,
+		Ref:  &repo.Branch,
+	}
+	for {
+		treeNode, resp, err := g.client.Repositories.ListTree(p.ID, &options)
```
rishabh625, line 128: can we include one above this listTree call into a loop.

rishabh625 (changes requested):

pkg/services/scm_provider/gitlab.go
dsch: Not sure if I understand you right: Do you like me to merge the duplicated `listTree` call into a loop?

If yes, I can't see how this will improve readability or maintainability of this function. Specially since the result of this two calls are differently interpreted:
1. if it's a directory any none empty list is ok
2. if it's a file we have to iterate over all pages to check if the directory contains the requested file

rishabh625: Yes i feel it's a duplicated code and hence should be merged.

since it's a function, function returns whether path exists or not, 

on first call if it's directory  and in loop u can compare total items and break and return from function,

Currently same call is done twice

I feel implementing in  for loop with comments will improve readability.

rishabh625: And thanks for addressing it,

I would request you to move this pr into Argo-cd repository as application set is merged there and future relase will happen along with argo-cd

rishabh625:
@dsch :  Can u please move this PR into argocd as applicationset is [moved](https://github.com/argoproj/argo-cd/pull/8864). If u find it laborious, i can do it on your behalf with your consent

dsch:
@rishabh625 I create a PR in argocd: https://github.com/argoproj/argo-cd/pull/8997
Can you please review it?

dsch:
Closing this PR, since https://github.com/argoproj/argo-cd/pull/8997 is merge into [argo-cd](https://github.com/argoproj/argo-cd).

PR #554, opened 2022-03-18
feat: removed ksonnet

gdsoumya:
Signed-off-by: Soumya Ghosh Dastidar <gdsoumya@gmail.com>

This PR removes ksonnet support, ref: https://github.com/argoproj/argo-cd/issues/5256

Changes:
1. sets the argo-cd pkg in go.mod to the current master [commit](https://github.com/argoproj/argo-cd/commit/30415e0b0b54d2a334dee330b42d7040508c1d5d)
2. removes ksonnet references from codebase


alexmt (changes requested):
We've agreed to prioritize merging the application set into Argo CD due to tight coupling between the projects. I believe this is already work in progress so I would suggest to work on merging instead.

@jgwest  WDYT?

gdsoumya:
> We've agreed to prioritize merging the application set into Argo CD due to tight coupling between the projects. I believe this is already work in progress so I would suggest to work on merging instead.
> 
> @jgwest WDYT?

Sure @alexmt, do we have independent subtasks tasks available for this? If not then I can collaborate on any exiting effort on the same.

rishabh625:
Thanks @gdsoumya , i am already taking care of this in migration [PR](https://github.com/argoproj/argo-cd/pull/8775)

rishabh625:
@wtam2018 we can close this is implemented in [#8864](https://github.com/argoproj/argo-cd/pull/8864) of argocd

rishabh625:
@gdsoumya : can we please close this?

PR #566, opened 2022-03-29
feat: New SCM and pull request generators for Gitea 

dmolik:
An initial try at adding Gitea support


dmolik:
Here's my first crack at supporting Gitea for the pull request and SCM generators. I haven't implemented API paging or Repo labels, I wanted to get some eyes on the PR before I went off and added more. Also the Gitea client doesn't support contexts well, so I didn't attempt to incorporate them, I can implement a work around if desired.

dmolik:
Closing via: https://github.com/argoproj/argo-cd/pull/8989

PR #575, opened 2022-04-09
docs(applicationset): update git generator example

Lemmah:
path[0] does not work with more than one application
path[0] ends up with all apps being called "example"

- tried copy pasting as the lazy a** I am and example on docs didn't work as expected.
- path.basename is easy thus very predictable.


rishabh625:
@Lemmah : Please add this changes in argoproj/argo-cd , as this code is been migrated there

Issue #446, opened 2021-12-28
argocd-applicationset-controller should run as non-root


haslersn:
argocd-applicationset-controller should run as non-root in order to be compatible with the Restricted policy of [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

PR #582, opened 2022-05-14
feat: run as non-root

sathieu:
Fixes: #446

Inspired by [argocd's `Dockerfile`](https://github.com/argoproj/argo-cd/blob/1186fee7dc3c1c38e1095fafc34f418dd607c361/Dockerfile#L43-L47).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,10 +24,17 @@ RUN make build
 FROM docker.io/library/ubuntu:21.10
 
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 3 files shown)


rishabh625:
@sathieu : This repository is no more maintained do check Readme

sathieu:
@rishabh625 Thanks fr the info.

it looks like this will be fixed in argocd 2.4.0 https://github.com/argoproj/argo-cd/commit/c77cf66aa1f7298216c91e9592f21bc5d924141c

Closing

PR #641, opened 2023-07-28
add a organizationNormalized template value for SCM provider

prune998:
From the issue https://github.com/argoproj/applicationset/issues/640

This MR adds a `organizationNormalized` parameter to the SCM provider to ensure we can then use the Org name in templates


prune998:
Tests are faining in the docker build image... which is using `golang:1.13.4`... so old :)

